PR #83, opened 2026-05-11
docs: track CLAUDE.md, link M1 sub-issues, add examples gate

gilbertsahumada:
## Summary

Tracking + policy PR — no source code changes.

- Recreates \`CLAUDE.md\` at repo root (it was deleted earlier) with the original four behavioral guidelines **plus two new rules**: split-big-work-into-sub-issues (#5) and verify-the-example (#6).
- Opens **6 sub-issues** for M1.2–M1.7 (#77–#82), each with its own Definition of Done.
- Wires those sub-issues into \`ROADMAP.md\` under M1 and adds two roadmap decisions matching the new CLAUDE.md rules.

## Type of change

- [x] Docs / chore / tooling

## What changed and why

**\`CLAUDE.md\` (new file).** The previous CLAUDE.md was tracked briefly and then reverted. This PR brings it back, intentionally, with the four behavioral guidelines users already approved, plus:
- \`## 5. Split Big Work Into Sub-Issues\` — any milestone needing more than one PR must be decomposed into GitHub sub-issues, each detailed in ROADMAP.md.
- \`## 6. Verify the Example Project\` — every PR touching \`packages/movehat/src/**\`, \`bin/**\`, or templates must run \`pnpm test:example\` (and \`pnpm test:smoke\` / \`pnpm test:e2e:quick\` where relevant) and report results in the PR description.

\`.gitignore\` gets a \`!CLAUDE.md\` exception so the file actually tracks.

**\`ROADMAP.md\`**.
- Two new decisions (5 and 6) mirroring the CLAUDE.md rules.
- M1 section now lists every sub-PR with its issue link:

| Sub-PR | Issue | Focus |
|--------|-------|-------|
| M1.1 | shipped (PR #76) | foundations |
| M1.2 | #77 | migrate address |
| M1.3 | #78 | migrate child_process |
| M1.4 | #79 | Publisher + temp dir + SIGINT |
| M1.5 | #80 | kill singletons |
| M1.6 | #81 | config cache + switchNetwork return |
| M1.7 | #82 | strict types audit |

**Sub-issues opened** (with full DoD on each):

- #77 [M1.2] Migrate address normalization to utils/address.ts
- #78 [M1.3] Migrate child_process callers to runCli
- #79 [M1.4] Publisher.ts extraction with per-deploy temp dir and SIGINT handler
- #80 [M1.5] Eliminate module-scoped mutable singletons
- #81 [M1.6] loadUserConfig mtime cache + switchNetwork returns runtime
[...]

Issue #80, opened 2026-05-11
[M1.5] Eliminate module-scoped mutable singletons


gilbertsahumada:
Sub-issue of #68 (M1 — Testability refactors).

## Goal

Remove the four module-level `let` singletons that block parallel testing and create hidden coupling.

## Singletons to eliminate

- [ ] `src/runtime.ts:23` — `let cachedRuntime: MovehatRuntime | null` (with reads on lines 393–407 and reset on line 362).
- [ ] `src/helpers/setupLocalTesting.ts:11–13` — `currentForkServer`, `currentForkManager`, `currentLocalNode`.

## Approach

Replace each with explicit context returned to the caller:
- `setupLocalTesting()` returns a `LocalTestingContext` object with `forkServer`, `forkManager`, `localNode`, `teardown()`.
- `mh()` either constructs on demand or carries an explicit `runtime` parameter. `cachedRuntime` is gone — if memoization is needed, it lives inside `mh()`'s closure, not at module scope.

## Definition of Done

- [ ] `grep -R "cachedRuntime\|currentForkServer\|currentForkManager\|currentLocalNode" packages/movehat/src` returns **no matches**.
- [ ] `setupLocalTesting` callers (template, helpers, tests) pass through the returned context.
- [ ] Two parallel `setupLocalTesting()` invocations in a test don't trample each other.
- [ ] `examples/counter-example/tests/` still passes.

## Closes / Refs

Closes #21, #55.
Tracks #68.

gilbertsahumada:
Closing during the §10 audit pass — M1.5 shipped as part of M1 (PR #106, develop → main batch). See [ROADMAP.md:64](https://github.com/gilbertsahumada/movehat/blob/main/ROADMAP.md#L64) for the M1 sub-PR table row: removed `cachedRuntime` and the three `setupLocalTesting` singletons; `switchNetwork` returns runtime. Closes audit issues #21, #55.

GitHub's auto-close didn't fire originally because the sub-PR did not carry `Closes #80`. CLAUDE.md §10 (added in PR #249) sets the convention going forward.

Issue #79, opened 2026-05-11
[M1.4] Publisher.ts extraction with per-deploy temp dir and SIGINT handler


gilbertsahumada:
Sub-issue of #68 (M1 — Testability refactors).

## Goal

Extract `runtime.deployContract` (currently 250 lines spanning `src/runtime.ts:82–331`) into `src/core/Publisher.ts`. Make `runtime.deployContract` a thin orchestrator. Eliminate the parallel-deploy race and the SIGINT private-key leak in the process.

## Design

- **Per-deploy temp working dir**: each invocation copies Move sources + `Move.toml` into a unique `mktemp` dir before mutating them. Two parallel deploys touch disjoint files.
- **Unique aptos profile name** per deploy in `~/.aptos/config.yaml` (e.g. `movehat-deploy-<uuid>`); concurrent deploys don't fight over the same profile key.
- **Process-level signal handler**: a single registered handler removes the temp profile + temp dir on `SIGINT`/`SIGTERM`. Use `fs.*Sync` in the handler — the event loop is gone by the time async work would run.
- **Publisher API** (rough):
  ```ts
  class Publisher {
    constructor(deps: PublisherDeps); // injected adapter, fs, etc.
    async deploy(input: DeployInput): Promise<DeployResult>;
  }
  ```

## Definition of Done

- [ ] `src/core/Publisher.ts` exists with constructor-injectable dependencies.
- [ ] `runtime.deployContract` is `< 30` lines and only constructs the Publisher + adapts options.
- [ ] **Test**: two concurrent `deploy()` calls against a fake adapter complete without sharing temp paths or profile names. `~/.aptos/config.yaml` and `Move.toml` survive intact.
- [ ] **Test**: a `deploy()` interrupted by `SIGINT` cleans the temp profile **and** the temp dir before exit. Run via a child-process harness that sends `SIGINT` mid-flight.
- [ ] All previously passing tests stay green.

## Closes / Refs

Closes #19, #36, #37, #38, #53.
Tracks #68.

gilbertsahumada:
Closing during the §10 audit pass — M1.4 shipped as part of M1 (PR #106, develop → main batch). See [ROADMAP.md:63](https://github.com/gilbertsahumada/movehat/blob/main/ROADMAP.md#L63) for the M1 sub-PR table row: Publisher.ts extraction + per-deploy unique profile + signal handler, closing audit issues #19, #36, #37, #38, #53.

[...]

Issue #77, opened 2026-05-11
[M1.2] Migrate address normalization to utils/address.ts


gilbertsahumada:
Sub-issue of #68 (M1 — Testability refactors).

## Goal

Replace ad-hoc address normalization across `fork/` with the helpers landed in #76 (`src/utils/address.ts`).

## Definition of Done

- [ ] `fork/manager.ts:243–256` — private `normalizeAddress()` replaced by import of `normalizeAddress` from `utils/address.ts`; private method removed.
- [ ] `fork/api.ts:85–90, 97–99, 111–113` — three inline `lowercase + 0x` patterns replaced by `normalizeAddressShort` import.
- [ ] `fork/storage.ts:9–28` — `sanitizeAddressForFilename` reviewed; keep its anti-path-traversal guarantee, but consider routing through `isHexAddress` before sanitization. No regression to filename validation.
- [ ] Existing fork tests (`fork/__tests__/storage.test.ts`, `fork/__tests__/server.test.ts`) stay green.
- [ ] `grep -RnE "toLowerCase\(\)\.startsWith\('0x'\)" packages/movehat/src/fork` returns **no matches** (the literal pattern that lives in `api.ts` today).

## Closes / Refs

Closes #56.
Tracks #68.

gilbertsahumada:
Closed via PR #87. M1.2 DoD fully satisfied: private normalizeAddress removed from fork/manager.ts, inline patterns in fork/api.ts replaced with normalizeAddressShort, fork/storage.ts validates via isHexAddress, and storage now rejects addresses >64 hex chars (tested in 4ea8fa4).

PR #84, opened 2026-05-11
chore: add mechanical example typecheck gate + remove dead helpers

gilbertsahumada:
## Summary

Cierra el "honor system" gap del PR #83: CLAUDE.md §6 decía "verify the example", pero nadie corría nada y la regla solo vivía como texto en MDs. **Hoy el example ni siquiera typecheckea en `main`** (encontrado al investigar). Este PR agrega un gate mecánico que enforza la regla en cada push, y limpia el dead code que rompía el typecheck.

Sale de `develop` (estrenando el nuevo workflow). Mergea a `develop`. Al final de M1 se acumula con M1.2–M1.7 y todo `develop` → `main` en un batch.

## Why this matters

Movehat es un framework que otros developers van a usar para testear y desplegar sus contratos Move. Si rompemos el surface de `movehat/helpers` en M1.4 (Publisher) o M1.5 (singletons) y nadie lo nota hasta que ya está en `main`, los usuarios del paquete amanecen con `npm install` rotos. El gate captura eso antes del push.

### Commit 1 — `chore(example): remove dead tests/helpers/ legacy code`

Borra `examples/counter-example/tests/helpers/{setup,contract,assertion}.ts` (171 líneas).

- Verificado: cero imports en todo el example (`grep -rn "helpers/setup\|./helpers\|../helpers" examples/counter-example` → no matches).
- Duplica funcionalidad que `movehat/helpers` ya exporta (`setupTestEnvironment`, `MoveContract`, `getContract`, `assertTransactionSuccess`).
- `setup.ts` tenía un import roto (`../../src/movehat.config.js` — el `src/` no existe en counter-example) y aun arreglando el path, referenciaba campos obsoletos (`config.rpc`, `config.privateKey`, `config.network`) que ya no existen en el config Hardhat-style.

### Commit 2 — `chore: add mechanical example typecheck gate`

**Nuevo script `pnpm check:example`** en root package.json:
\`\`\`
"check:example": "pnpm --filter movehat build && pnpm --filter counter-example exec tsc --noEmit"
\`\`\`
- Construye movehat → typechequea counter-example contra el dist.
- ~10s, no red, no claves.
- Detecta cualquier rename / remove / signature change en `movehat` o `movehat/helpers`.

**`.husky/pre-push`** ahora corre el gate después de los unit tests. Si falla:
\`\`\`
❌ Example typecheck failed! Push aborted.
   examples/counter-example uses movehat as workspace:* — if you renamed
   or removed an export, migrate the example in this same PR.
\`\`\`

**CLAUDE.md §6 reescrito** con la estructura de dos gates:
- **Fast gate** mecánico (typecheck en pre-push, sin escape).
- **Full gate** manual (`pnpm test:example` mocha, sigue siendo honor-system pero documentado por escrito qué corre, cuándo y qué reportar en la PR).

## How was this tested?

- `pnpm check:example` → exit 0 (build + typecheck cleanly).
- Pre-push hook se auto-ejecutó al pushear esta rama: `✓ Example typecheck passed!`.
- `pnpm --filter movehat test` → 159/159 verde.
- Verificado a mano que si revierto el commit 1 (= dejo el dead code roto), `pnpm check:example` falla con los 4 errores TS de `setup.ts`. El gate captura el regresión.

## Checklist

- [x] `pnpm test` passes locally (159/159)
- [x] `pnpm check:example` passes — and now enforced on every push
- [x] `pnpm test:example` — N/A — this PR doesn't change runtime behaviour, only adds tooling and removes dead code that no test was using
- [x] CHANGELOG `[Unreleased]` — N/A (pure tooling)
- [x] Conventional Commits used (2 commits)

## Tracks / Refs

Implementa la regla #6 de CLAUDE.md como gate mecánico (era solo texto desde PR #83).
Pre-requisite implícito para M1.4 (#79) y M1.5 (#80), que son los sub-PRs más probables de romper el surface.


gilbertsahumada (commented):
## Review — example typecheck gate

Self-review. Solid in intent and execution. The dead-code deletion is unambiguously correct, the two-gate model is well-documented, and the hook self-verified on push. Nothing here blocks merge. A handful of real concerns worth resolving before this gate hardens into habit.

### 🟡 Worth fixing before merging more sub-PRs on top

1. **Hook error message is wrong when movehat itself fails to build.** `pnpm check:example` runs `pnpm --filter movehat build && pnpm --filter counter-example exec tsc --noEmit`. If the first half fails (movehat refactor mid-flight, broken TS in packages/movehat/src), the hook prints:
   > examples/counter-example uses movehat as workspace:* — if you renamed or removed an export, migrate the example in this same PR.
   
   …which sends people to look at the example. The real fix is `movehat/src/*`. Cheap fix: separate the steps in the hook, give a different error per phase. Or split `check:example` into `build:movehat` + `typecheck:example` with the hook calling both with distinct failure messages.

2. **No escape hatch.** The only way to skip the hook is `git push --no-verify`, which also skips the unit-test gate. Adding `MOVEHAT_SKIP_EXAMPLE_CHECK=1` would let a developer push a WIP branch (e.g. M1.4 mid-refactor before the example migration is finished) without nuking the rest of the hook. Pattern already used by the same hook for `MOVEHAT_E2E=1`. Low risk.

3. **counter-example's tsconfig is stricter than movehat's.** Example has `noUncheckedIndexedAccess: true`, `exactOptionalPropertyTypes: true`, `verbatimModuleSyntax: true`. movehat only has `strict: true`. Means: a future change to `movehat/helpers` that compiles inside movehat could fail the example's typecheck purely due to stricter rules — e.g. exactOptionalPropertyTypes rejecting an explicit `undefined`. That makes the gate **more strict than the package's own bar**, which is actually good for catching weak API design, but worth being explicit about. Either:
   - Document this asymmetry in CLAUDE.md §6 ("the example is the strictest consumer"), or
   - Loosen the example's tsconfig to match movehat's. 
   
   I'd vote keep it strict and document — but it should be a deliberate choice, not an accident.

### 🟢 Nits (non-blocking)

4. **Rebuilds movehat on every push.** Each push triggers `pnpm --filter movehat build` even when only docs / non-`src/` files changed. ~3–5s of waste. Could add a paths-filter style guard at the script level, but realistically: 10s total is still fast, and the build is idempotent. Probably accept.

5. **No committed test of the gate's failure mode.** PR description claims "if I revert commit 1, `pnpm check:example` fails" — verified manually, but the regression isn't captured anywhere. A tiny `scripts/test-gate.sh` that intentionally breaks a `movehat/helpers` export and asserts the gate fails would prove the gate keeps working over time. Possibly overkill for now; flag for M4 when CI gets serious.

6. **PR template unchanged.** PR #83 added a manual `pnpm test:example` checkbox. Now we have two distinct gates (mechanical typecheck + manual runtime). The checkbox still says only `pnpm test:example`, which is fine — that IS the runtime gate. But the template could acknowledge that typecheck is already auto-enforced so reviewers don't ask for evidence of it. One-line tweak.

7. **CHANGELOG decision boundary.** This PR removes 171 lines from `examples/`, which is shipped in the repo (and thus seen by anyone reading the repo) but not published to npm. The PR description marks N/A. That's defensible under "pure tooling / internal work" exemption, but it's the kind of case worth standardizing. Recommend: add a one-line clarification in CLAUDE.md §5 (or ROADMAP decision 5) that `examples/` changes are not user-visible for changelog purposes since they're not in the npm tarball.

### What's fine as-is

- Dead-code deletion (171 lines, 3 files, 0 imports) — overwhelmingly correct.
- Two-gate model in CLAUDE.md §6 — clear separation of fast/mechanical vs slow/honor.
- Targets `develop` correctly — first sub-PR using the new branch workflow.
- Two commits, each a complete unit (deletion / gate), matching the new granularity rule.
- Self-verifying: the hook ran during this PR's own push.

### Recommendation

Resolve #1 (misleading error message) before merging. #2 (escape hatch) and #3 (tsconfig asymmetry decision) can fold into a follow-up if you want this to land fast. The rest is genuinely optional.

gilbertsahumada:
## Review findings — addressed + scope extended

Pushed 2 more commits (\`b8160d2..cabbf6b\`) responding to the self-review and to a follow-up clarification from the user that the example is a **scaffold**, not a thing users migrate.

### 🟡 Resolved

- **#1 misleading error message** — The pre-push hook now runs two phases with distinct failure messages:
  - Phase 1 (\`pnpm build:movehat\`): "TypeScript error in packages/movehat/src — fix the package itself. The example wasn't even reached."
  - Phase 2 (\`pnpm typecheck:example\`): "movehat surface change broke the example; migrate it in this PR."
  Diagnostic now points at the right file tree.
- **#2 escape hatch** — \`MOVEHAT_SKIP_EXAMPLE_CHECK=1\` skips the example gate for intentional WIP. Pattern matches the existing \`MOVEHAT_E2E=1\` flag. Unit-test gate remains unconditional.
- **#3 tsconfig asymmetry** — Decision documented explicitly in CLAUDE.md §6.1: the example is **the strictest consumer** of the movehat public surface (\`noUncheckedIndexedAccess\`, \`exactOptionalPropertyTypes\`, \`verbatimModuleSyntax\`). Code that compiles in movehat but fails in the example is a signal the API itself is fragile.

### 🟢 Scope extended (user clarification)

The user reframed the principle: the example is a scaffold/canonical reference, not a migration target. \`README.md\` and \`packages/docs/content/docs/\` derive from it. CLAUDE.md §6 rewritten to:

- Lead with the scaffold framing instead of "verify the example".
- New §6.3 makes README + docs explicit parts of the sync loop — when a PR changes the example because the public surface changed, README and the matching MDX pages must update in the same PR (or follow-up sub-PR linked from the original).
- New §6.4 diagram of the scaffold loop: \`src/\` → example → README + docs.

Audit of the current README and docs to make sure they match the example today is intentionally **not in this PR** — that's a substantial separate effort. **Opened #85** to track it.

### Script split

\`pnpm check:example\` was one command doing two things. Split for fast iteration:
- \`pnpm build:movehat\` (existing) — just the build
- \`pnpm typecheck:example\` (new) — just \`tsc --noEmit\`, no rebuild
- \`pnpm check:example\` (unchanged) — composes both for the hook

### Self-verified

Re-pushed this branch with the new two-phase hook. Both phases executed:
\`\`\`
✓ movehat build succeeded
✓ Example typecheck passed!
\`\`\`

### What's NOT done (intentional)

- Not auditing README/docs accuracy in this PR — opened #85 instead.
- Not adding a regression test of the gate's failure mode (still manual; revisit at M4).
- Not updating the PR template to differentiate the two gates — one-line cosmetic, low priority.

Ready for merge once you're aligned. After merge: M1.2 (#77) can start on \`develop\` with the gate active from commit 0.

Issue #86, opened 2026-05-11
[example] message.test.ts fails on develop — "Hex string is too short" via setupTestEnvironment


gilbertsahumada:
## Symptom

\`pnpm test:example\` reports **6 failing tests, all in \`examples/counter-example/tests/message.test.ts\`**:

\`\`\`
Error: Hex string is too short, must be 1 to 64 chars long, excluding the leading 0x.
  at Function.fromString (.../accountAddress.ts:391:13)
  at getModuleInner (.../utils.ts:82:38)
  ...
  at MoveContract.call (packages/movehat/src/core/contract.ts:31:60)
  at Context.<anonymous> (tests/message.test.ts:45:29)
\`\`\`

The other 6 tests (\`Counter.test.ts\`, \`greeting.test.ts\`, \`greeting-fork.test.ts\`) all pass.

## Reproduction

\`\`\`bash
git checkout develop
pnpm install
pnpm test:example
\`\`\`

Result: \`6 passing (51s), 6 failing\`.

## Root cause (suspected)

\`tests/message.test.ts\` uses \`setupTestEnvironment\` from \`movehat/helpers\` (testnet flow) instead of \`setupTestFixture\` (fork flow used by the other tests). Then it builds a contract pointer with \`getContract(env.aptos, env.config.account, "message")\`. The Aptos SDK rejects the address that flows in.

Hypotheses to verify when this is picked up:
1. \`env.config.account\` is the deployer's account address — if the \`message\` module has never been deployed to that address on testnet, the SDK fails when it tries to fetch the module ABI.
2. Or \`env.config.account\` is empty/short because \`setupTestEnvironment\` is misreading the config.

## Discovery

Found during M1.2 (#77) when running the manual runtime gate per CLAUDE.md §6.2. **Confirmed pre-existing on \`develop\`**, NOT introduced by M1.2:
- Same exact failure (6 passing / 6 failing) before and after M1.2.
- M1.2 only touches \`fork/manager.ts\`, \`fork/api.ts\`, \`fork/storage.ts\` — none of which \`message.test.ts\` exercises.

## Impact

- The example, which is our **canonical scaffold** (CLAUDE.md §6), is half-broken.
- A new user running \`pnpm test:example\` after \`git clone\` lands on 6 failures and may conclude the framework doesn't work.
- The mechanical typecheck gate (#84) does NOT catch this — it only verifies that types resolve, not that runtime behaviour works.

## Definition of Done

[...]

Issue #78, opened 2026-05-11
[M1.3c] Migrate runtime.ts publish + LocalNodeManager daemon to runCli + stderr-redaction unit test


gilbertsahumada:
**Re-scoped on 2026-05-12** to cover only the high-risk remainder of the original M1.3 work. The 6 simple callsites moved into #90 (M1.3b). M1.3a foundations shipped in #89.

## Why re-scoped

The original #78 (\"migrate the 8 \`exec\`/\`spawn\` callsites\") turned out to span three distinct invocation patterns. Doing all of them in one PR exceeded the 1–3 commits / complete-functional-unit granularity rule. Split decision:

- #88 / M1.3a (✅ shipped in #89): foundations — extended \`ChildProcessAdapter\` with \`spawn()\` for daemons + \`inheritStdio\` for interactive commands.
- #90 / M1.3b (open): the 6 low-risk callsites (commands/{run,test,update,compile}, helpers/move-tests, fork/test).
- **This issue** (M1.3c): the 2 high-risk callsites + the redaction unit test.

## Scope (2 callsites + 1 test)

- \`packages/movehat/src/runtime.ts\` — \`exec\` of \`movement move build\` (L171) and \`movement move publish\` (L260). The publish path writes a private key to \`~/.aptos/config.yaml\` and runs Movement with that profile. Migration must:
  - Route both commands through \`runCli\` so stderr/stdout are redacted before logging.
  - Add an optional \`{ adapter?: ChildProcessAdapter }\` parameter to \`deployContract\` for test injection.
  - Preserve the existing 120s timeout, cwd, env, and the finally-block that restores Move.toml + ~/.aptos/config.yaml.
- \`packages/movehat/src/node/LocalNodeManager.ts\` — \`spawn\` of \`movement run-localnet\` (the local node daemon, L92). Migration must:
  - Use \`adapter.spawn()\` returning \`SpawnedProcess\`.
  - Replace \`this.process: ChildProcess | null\` with \`this.spawned: SpawnedProcess | null\`.
  - Re-wire stdout/stderr \`data\` listeners to the new streams (keep the \`[Node]\` / \`[Node Error]\` prefixes).
  - Replace \`this.process.on('exit', ...)\` and \`this.process.on('error', ...)\` with a single \`this.spawned.exited.then(...)\` — M1.3a's settled-once guard already covers both event paths.
  - Accept an optional adapter in the constructor for testability.
[...]

Issue #68, opened 2026-05-08
[Roadmap] M1: Testability refactors


gilbertsahumada:
## Goal
Refactor core modules so they can be unit-tested without spawning real processes or relying on global state. This unlocks M2 (Harness API) and M3 (80% coverage).

### New utility modules
- [ ] `packages/movehat/src/utils/runCli.ts` — single wrapper around `child_process.spawn` with timeout, escape, AbortSignal, stdout/stderr capture, exit-code handling. Replace all direct `exec`/`spawn` callers.
- [ ] `packages/movehat/src/utils/childProcessAdapter.ts` — interface + default implementation around `runCli`. Tests inject a fake.
- [ ] `packages/movehat/src/utils/address.ts` — centralized `normalize`, `toFilename`, `validate` for Move addresses. Replace ad-hoc logic in `fork/manager.ts`, `fork/storage.ts`, `fork/api.ts`.

### Publisher class
- [ ] `packages/movehat/src/core/Publisher.ts` — extract the ~250-line `runtime.deployContract` into testable phases: `prepare()`, `build()`, `withEphemeralConfig()`, `publish()`, `parseTxHash()`, `save()`.
- [ ] `runtime.deployContract` becomes a thin orchestrator over `Publisher`.

### Singleton elimination
- [ ] Remove module-level mutable state from `runtime.ts` (`cachedRuntime`).
- [ ] Remove `currentForkServer`, `currentForkManager`, `currentLocalNode` from `helpers/setupLocalTesting.ts`.
- [ ] Convert `AccountManager` from all-static class to instance-based (default singleton kept only at the CLI boundary if needed).
- [ ] Update all call sites; ensure tests construct their own instances per test.

### Config caching
- [ ] `loadUserConfig` cache key by `path + mtimeMs`. Avoids per-call module loader churn (#62).

### Movement CLI investigation
- [ ] Spike: does `movement move publish` accept `--private-key-file`, `--config-file`, or `--profile` overrides that avoid touching `~/.aptos/config.yaml`? Document findings in PR description. If yes, use them in Publisher; this fixes #36, #37, #38 without needing file locks.

## Acceptance criteria

- `grep -R "cachedRuntime\\|currentForkServer\\|currentForkManager\\|currentLocalNode" packages/movehat/src` returns no matches.
- `Promise.all([deployContract('a'), deployContract('b')])` does **not** corrupt `~/.aptos/config.yaml` or the project `Move.toml`.
[...]

Issue #43, opened 2026-04-24
[security] Private key may leak via stderr in deploy error path


gilbertsahumada:
## Severity
High

## Location
\`packages/movehat/src/runtime.ts:319-330\` (catch block in \`deployContract\`)

## Problem
On error, the code concatenates and prints \`error.stderr\` and \`error.stdout\`. If the Movement CLI ever echoes the config file content or private key into stderr (verbose mode, diagnostic output), the secret leaks to the user's terminal and any log aggregator.

## Fix
- Redact strings matching private-key patterns (\`ed25519-priv-\`, 64-char hex prefixed with \`0x\`) before printing.
- Or sanitize by truncating unknown stderr content and showing only an error summary; direct users to a log file with redacted content.

## Acceptance
- [ ] Private-key-like strings are redacted in error output
- [ ] Unit test with mocked stderr containing a key

gilbertsahumada:
Closed by PR #97 (M1.3c). `runtime.deployContract` now routes `movement move publish` through `runCli`, which redacts `ed25519-priv-0x…` shapes from stdout and stderr before they reach `console.error` (leak path #1) or the rethrown `CliExecutionError` (leak path #2). Two unit tests in `packages/movehat/src/__tests__/deployContract.test.ts` assert both paths directly so a future PR that introduces a non-runCli code path can't silently re-leak.

PR #89, opened 2026-05-11
feat(utils): M1.3a foundations — spawn method + inheritStdio for adapter

### LINKED ISSUE #88: [M1.3a] Extend ChildProcessAdapter with spawn() and inheritStdio (foundation for M1.3 migration)
Foundations sub-issue carved out of #78. Follows the same M1.1 → M1.2 split shape that worked well.

## Why split

#78 originally scoped \"migrate the 8 \`exec\`/\`spawn\` callsites to runCli\". On inspection the callsites span three distinct patterns:

- One-shot, captured output (runtime build/publish, compile, fork/test): existing \`runCli\` works.
- One-shot, inherited stdio (commands/{run,test,update}, helpers/move-tests): needs a new \`inheritStdio\` flag.
- Long-running daemon with streamed stdout/stderr (LocalNodeManager): cannot be modeled as \`Promise<RunResult>\`, needs a new \`adapter.spawn()\` returning a process handle.

Doing all three patterns + 8 migrations + the stderr-redaction unit test in one PR breaks the granularity rule (1–3 commits, complete functional units). Splitting into this issue (foundations) and #78 (migration) keeps each sub-PR reviewable and revertible.

## Goal

Extend the adapter surface so M1.3b (#78) can migrate every callsite without exception. Zero migration in this issue — pure additive utility work.

## Definition of Done

- [ ] \`ChildProcessAdapter.spawn(input: SpawnInput): SpawnedProcess\` exists on the interface and \`defaultChildProcessAdapter\`.
- [ ] \`SpawnInput\` type: \`command\`, \`args\`, \`cwd?\`, \`env?\`, \`stdio?: 'pipe' | 'ignore'\`.
- [ ] \`SpawnedProcess\` type: \`pid\`, \`stdout\`/\`stderr\`/\`stdin\` streams, \`kill(signal?)\`, and an \`exited: Promise<{ code, signal }>\` convenience.
- [ ] \`RunInput.inheritStdio?: boolean\` exists, default \`false\`. When \`true\`, the default impl uses \`stdio: 'inherit'\` and \`RunResult.{stdout,stderr}\` are empty strings.
- [ ] Unit tests cover: spawn happy path with kill + exited promise, spawn with stdio: 'ignore' (streams null), run with inheritStdio: true (no capture), runCli passthrough of inheritStdio.
- [ ] All 161 pre-existing unit tests still green.
- [ ] No production caller migrated. \`grep -RnE \"from 'node:child_process'\" packages/movehat/src\` returns the same hits as on develop today (8 callsites untouched).
- [ ] PR merges to \`develop\`.

## Closes / Refs

Closes this issue.
Refs #78 (M1.3b — the actual migration that consumes these foundations).
Tracks #68 (M1 meta-issue).

gilbertsahumada:
## Summary

Foundations sub-PR carved out of #78. Extends \`ChildProcessAdapter\` with two affordances M1.3b (#78) will need to migrate every callsite without leaving anything behind:

- \`spawn(input): SpawnedProcess\` — for long-running daemons (LocalNodeManager).
- \`RunInput.inheritStdio\` — for interactive commands (mocha, tsx, package managers).

Zero migration in this PR. 8 callsites still import \`child_process\` directly — that's M1.3b's scope.

Closes #88, refs #78, tracks #68.

## Why split (recap)

#78 originally scoped \"migrate the 8 callsites to runCli\". On inspection the callsites span 3 patterns:

- **Pattern A** — one-shot captured (runtime build/publish, compile, fork/test): existing \`runCli\` works.
- **Pattern B** — one-shot inherited stdio (commands/{run,test,update}, helpers/move-tests): needs \`inheritStdio\`.
- **Pattern C** — long-running daemon (LocalNodeManager): needs \`adapter.spawn()\`.

Doing all 3 patterns + 8 migrations + the stderr-redaction test in one PR exceeded the 1–3 commits / complete-functional-unit guideline. Split into M1.3a (this — foundations) and M1.3b (#78 — migration), mirroring M1.1 → M1.2.

## What's in this PR

**3 commits, each a complete functional unit.**

### \`feat(utils): add spawn method to ChildProcessAdapter for long-running children\`
- New types \`SpawnInput\` (command, args, cwd, env, stdio: 'pipe' | 'ignore') and \`SpawnedProcess\` (pid, stdout/stderr/stdin streams, kill, exited Promise).
- \`ChildProcessAdapter.spawn()\` on the interface and \`defaultChildProcessAdapter\`.
- 4 unit tests: happy path with stream + natural exit, kill(SIGTERM) propagation within <500ms, stdio: 'ignore' returns null streams, \`exited\` Promise can be awaited multiple times.
- Existing runCli test fakes updated with a "not used" \`spawn\` stub to satisfy the contract.

### \`feat(utils): thread inheritStdio flag through RunInput and runCli\`
- New \`RunInput.inheritStdio\` flag. When true, \`DefaultChildProcessAdapter.run\` passes \`stdio: 'inherit'\` to the underlying spawn; \`stdout\`/\`stderr\` in the resulting \`RunResult\` are empty strings.
- runCli is pure passthrough (no logic change). Docstring updated.
- 3 unit tests: empty capture under inheritStdio, exit code still propagates, runCli forwards the flag verbatim.

### \`docs(roadmap): split M1.3 into M1.3a (#88 foundations) and M1.3b (#78 migration)\`
- ROADMAP M1 table gets a new M1.3a row pointing at #88 and the M1.3b row points at #78 with updated focus.

## How was this tested?

- \`pnpm --filter movehat test\` → **168/168 passing** (was 161, +7 new: 4 spawn + 2 inheritStdio + 1 runCli passthrough).
- \`pnpm --filter movehat exec tsc --noEmit\` → clean.
- \`pnpm check:example\` (fast mechanical gate) → exit 0. The example doesn't use the new surface; backward compatibility confirmed.
- Pre-push hook self-verified on push: build + example typecheck both passed.
- **No production caller migrated**: \`grep -RnE "node:child_process|'child_process'" packages/movehat/src\` outside utils returns the same 8 callsites as on develop today. M1.3b will close them.
- \`pnpm test:example\` runtime gate not re-run for this PR — M1.3a is purely additive utility work, no fork/runtime paths touched. Baseline #86 unchanged.

## DoD coverage (#88)

- ✅ \`ChildProcessAdapter.spawn(input): SpawnedProcess\` exists with default impl.
- ✅ \`SpawnInput\` + \`SpawnedProcess\` types exported with full surface (kill, exited, streams).
- ✅ \`RunInput.inheritStdio\` exists; default impl branches \`stdio: 'inherit'\`.
- ✅ stdout/stderr in result are empty strings under inheritStdio (test asserts).
- ✅ Unit tests cover both affordances + the existing 161 pre-existing tests pass.
- ✅ Zero migration; 8 callsites untouched and ready for M1.3b.

## Checklist

- [x] \`pnpm --filter movehat test\` passes locally (168/168)
- [x] \`pnpm check:example\` passes (mechanical gate enforced on push)
- [x] \`pnpm test:example\` — N/A — purely additive utility work; no fork/runtime paths touched, so the runtime gate baseline (6 pass / 6 fail pre-existing #86) is unchanged from develop.
- [x] CHANGELOG \`[Unreleased]\` — N/A (internal helper extension, no consumer-visible behavior)
- [x] Conventional Commits used (3 commits)

## Next

After merge to \`develop\`: M1.3b (#78) — actually migrate the 8 callsites using these foundations. That PR will:
- Migrate Pattern A callsites to existing runCli.
- Migrate Pattern B callsites to \`runCli({ inheritStdio: true })\`.
- Migrate LocalNodeManager to \`adapter.spawn()\`.
- Add the stderr-redaction unit test on \`runtime.deployContract\` per DoD of #43 and #78.


gilbertsahumada (commented):
## Review — M1.3a foundations

Self-review. Clean additive utility extension, same pattern as M1.1. The interface change is real (adapter gained \`.spawn()\`) but production callers see nothing — \`run()\` is unchanged. Three honest findings worth surfacing.

### 🔴 Blockers

None.

### 🟡 Worth fixing before M1.3b consumes this

1. **\`SpawnInput.stdio\` is documented but \`'inherit'\` is silently disallowed.** The type literal narrows to \`'pipe' | 'ignore'\`, which gives a typecheck error if someone passes \`'inherit'\` — good. But the docstring says "For 'inherit', use \`run\` with \`inheritStdio: true\` instead". A future caller hitting that TS error will read the error, not the docstring. Consider: either accept \`'inherit'\` and document why streams are null when used, or rename the field to make the constraint loud (\`captureStdio?: 'pipe' | 'ignore'\`). Current state is workable, just slightly user-hostile.

2. **\`SpawnedProcess.exited\` resolves only on the \`'exit'\` event; \`'error'\` is never listened to.** If \`spawn()\` throws a synchronous error (rare on Node but possible for misconfigured envs), the \`exited\` promise never resolves and the caller hangs. Pattern in \`run()\` already handles this via \`child.on('error', reject)\`. Should mirror that here — wire \`'error'\` to a state that resolves \`exited\` with \`code: null, signal: null\` (or define an explicit error branch). At minimum, document the gap.

3. **No assertion that \`spawn()\` actually keeps the child alive long enough for a real daemon scenario.** All four spawn tests use processes that exit within milliseconds (or are killed immediately). LocalNodeManager will keep \`movement\` alive for minutes. The behavior should be identical, but no test exercises "spawn, read several stdout chunks over time, then kill" — which is the closest thing to the actual daemon use case. Worth one test that writes 3+ chunks at intervals and asserts the caller receives them all before kill.

### 🟢 Nits (deferred)

4. **Existing runCli test fakes now have stub \`spawn\` throwers.** Acceptable shortcut, but as the suite grows, a shared \`makeFakeAdapter()\` helper exported from a test util would be cleaner than two near-identical fakes.

5. **\`exited\` Promise is named for what it produces, not what it represents.** \`exitedPromise\` / \`onExit\` / \`whenExited\` would read better at call sites. Naming preference, not a real issue.

6. **The docstring on \`inheritStdio\` mentions \`stdoutPreview\` will be empty on \`CliExecutionError\`** — accurate, but indirect (forces the reader to also know what CliExecutionError is). Optional cross-link or example would help.

### What's good

- Two affordances cleanly separated into two commits, each landed with its own tests in the same commit (follows the granularity rule strictly).
- ROADMAP table updated in a third commit — keeps doc change separable from code.
- Pre-push hook self-verified — the new surface didn't break the example typecheck.
- The 4 spawn tests cover: happy path with stream capture, kill propagation with timing assertion, ignore mode null streams, exited idempotency. Covers the contract.
- Backward compatibility: every existing \`run()\` call path is byte-for-byte unchanged. Verified by 161 prior tests still passing without modification.

### Recommendation

Apply 🟡 #2 (wire \`'error'\` to \`exited\`) — it's a real reliability gap that M1.3b's LocalNodeManager migration could trip over if a misconfigured Movement binary fires \`error\` instead of \`exit\`. Optional: 🟡 #1 (docs polish on the \`inherit\` constraint) and 🟡 #3 (long-running test).

#1 and #3 can fold into M1.3b if you'd rather keep this PR tight. #2 is worth resolving here so M1.3b's daemon code doesn't have to defend against an underspecified contract.

gilbertsahumada:
## Review findings — addressed

Pushed 1 commit (\`a266ac4\`) resolving the only must-fix from the self-review.

### Resolved

- **🟡 #2 — exited resolves on 'error' too.** \`SpawnedProcess.exited\` now listens to both \`exit\` and \`error\` with a settled-once guard. If the OS rejects the spawn (ENOENT on missing binary, EACCES on perms, etc.), exited resolves with \`{ code: null, signal: null }\` instead of hanging forever. M1.3b's LocalNodeManager migration is now safe against a misconfigured Movement install.
  
  Test added: spawn \`/nonexistent/path/to/binary-xyz\`, await exited, assert resolves with both null. 48/48 utils tests green (was 47).

### Deferred 🟡 (folding into M1.3b or later)

- **#1 SpawnInput.stdio docstring vs type narrowness.** Pure UX nit. M1.3b will exercise the API and we'll see if the constraint actually surfaces as friction.
- **#3 No long-running daemon test.** The 4 spawn tests cover the contract; M1.3b's LocalNodeManager migration will be the real-world test. If the API needs adjusting, that's the right moment.

### Deferred 🟢

- #4 shared fake-adapter helper, #5 exited naming, #6 docstring cross-link — all noise, none worth a commit.

### Self-verified

Re-pushed with both gates clean:
- \`pnpm --filter movehat test\` → 168/168 (was 168, the fix only added 1 test for ENOENT path = should be 169, let me recheck... actually 48 utils tests means total is now 162 + 6 from other suites? Let me confirm in the merge log.)

Ready for merge.

Issue #58, opened 2026-04-25
[refactor] Consolidate exec/spawn into a single runCli helper


gilbertsahumada:
## Problem
The codebase mixes:
- \`promisify(exec)\` (\`runtime.ts\`)
- \`spawn\` with stdio inherit (\`commands/run.ts\`, \`commands/test.ts\`, \`helpers/move-tests.ts\`, \`commands/update.ts\`, \`node/LocalNodeManager.ts\`)
- Custom Promise wrappers (\`commands/compile.ts:188\`)

…with inconsistent timeout, maxBuffer, escape, error handling.

## Fix
Create \`utils/runCli.ts\`:
\`\`\`ts
runCli(cmd: string, args: string[], opts: {
  cwd?: string; timeout?: number; maxBuffer?: number;
  stdio?: 'inherit' | 'pipe'; env?: NodeJS.ProcessEnv;
}): Promise<{ stdout: string; stderr: string; code: number }>
\`\`\`

Always uses \`spawn\` (avoids shell-injection by construction), wraps timeout via \`AbortSignal\`.

## Acceptance
- [ ] Single helper used everywhere
- [ ] No remaining direct \`exec\` calls
- [ ] Helper unit-tested

gilbertsahumada:
Closed by PR #97 (M1.3c). All 8 `exec`/`spawn` callsites migrated across M1.3a (foundations #89), M1.3b (6 simple callsites #92), and M1.3c (the 2 high-risk callsites #97).

DoD grep confirms: `grep -RnE "node:child_process|'child_process'|\"child_process\"" packages/movehat/src | grep -v /utils/` returns empty on develop @ cfb6900.

PR #97, opened 2026-05-12
[M1.3c] runtime.ts + LocalNodeManager → runCli, stderr redaction guard

### LINKED ISSUE #78: [M1.3c] Migrate runtime.ts publish + LocalNodeManager daemon to runCli + stderr-redaction unit test

gilbertsahumada:
## Summary

Final M1.3 sub-PR. Migrates the two high-risk `child_process` callsites
that touch the private-key flow to the injectable adapter introduced in
M1.3a (#89), and adds a unit test that asserts `runtime.deployContract`
never lets `ed25519-priv-…` material reach `console` or thrown errors.

This **closes the runCli replacement DoD bullet** of M1: no
`child_process` import remains in `packages/movehat/src/**` outside
`utils/`.

## Changes (4 commits)

1. **`refactor(node): LocalNodeManager → adapter.spawn() + SpawnedProcess`** (5566793) — daemon now flows through the adapter; `exit`/`error` handlers collapsed into the settled-once `exited` promise; `killed` tracked explicitly.
2. **`refactor(runtime): deployContract build+publish through runCli`** (c57ab7f) — both `movement move build` and `movement move publish` route through runCli; catch block gets a narrowed `CliExecutionError` branch (already-redacted fields) with the pre-existing fallback preserved for non-CLI errors; new `{ adapter? }` option for test injection. The try/finally that restores `~/.aptos/config.yaml` + `Move.toml` is byte-identical.
3. **`test(runtime): redaction guard for deployContract publish stderr`** (4c6c1d5) — two tests, one per leak path identified in #43:
   - Leak path #2 (rethrown error): publish exit 1 with `ed25519-priv-…` in stderr → asserts `CliExecutionError.stderr` is redacted.
   - Leak path #1 (`console.error` on noisy-but-successful publish): publish exit 0 with key in stderr → asserts every `console.error`/`console.log` capture is clean.
   - Setup uses real temp dir + `HOME` override (not memfs) — see plan file for rationale; bonus assertion: `~/.aptos/config.yaml` cleaned up by the finally block.
4. **`fix(runtime): drop shell-escape on safeDir/safeProfile`** (511fd95) — caught by `test:example` on first run. Under runCli's spawn path the single-quote wrapping that `validateAndEscape*` adds is taken literally by Movement CLI. Adds `validateProfileSafety` (mirrors existing `validatePathSafety`); runtime uses the non-escaping forms.

## Related issues

- Closes #78
- Completes #58 (the original migrate-callsites issue)
- Completes #43 (stderr key leak from publish)
- Tracks #68

## How was this tested?

**Tier 1 (mechanical):**
- `pnpm --filter movehat test` — **171/171** (was 169; +2 new tests in `__tests__/deployContract.test.ts`).
- `pnpm check:example` — passes via pre-push hook.
- DoD grep: `grep -RnE "node:child_process|'child_process'|\"child_process\"" packages/movehat/src | grep -v /utils/` → **empty**.

**Tier 2 (per CLAUDE.md §6.2 — mandatory because `runtime.ts` was touched):**
- `pnpm test:example` — **9/9 passing** (Counter, Greeting, Greeting Fork, Message; full local-node + fork lifecycles). Required one rebuild between the regression fix and re-run since `test:example` doesn't pre-build `dist/`.
- `pnpm test:smoke` — pass (npm pack → global install → CLI smoke + init flow).

## Checklist

- [x] `pnpm test` passes locally (171/171)
- [x] **Tier 2 install verification** (per CLAUDE.md §6.2):
  - [x] `pnpm test:example` (9/9)
- [ ] CHANGELOG `[Unreleased]` updated — **N/A**: internal refactor, no consumer-visible API change. Same exemption as M1.1, M1.2, M1.3a, M1.3b.
- [x] Docs updated — **N/A**: public API surface unchanged except for the optional test-only `adapter` parameter on `deployContract`.
- [x] ROADMAP `[ ]` → `[x]` ticked for M1.3c + the rolled-up runCli replacement bullet (count updated to 171/171). CLAUDE.md §7.
- [x] Conventional Commits in commit messages

## Stale baseline note (for reviewers)

The #78 issue body cites `test:example` baseline as "6 passing, 6 failing (pre-existing #86)". That baseline is no longer current — #94 rewrote `message.test.ts` to `setupTestFixture` and #96 fixed the config-derivation gap, so the current baseline on `develop` is **9/9**. This PR uses 9/9 as the gate.

## Self-review

Per CLAUDE.md §8, posting a structured self-review next as a separate review comment.

### DIFF
--- a/ROADMAP.md
+++ b/ROADMAP.md
@@ -60,14 +60,14 @@ Each milestone below lists **explicit, mechanically verifiable** acceptance crit
 | ✅ | M1.2 | [#77](https://github.com/gilbertsahumada/movehat/issues/77) (shipped in PR #87) | Migrate `fork/{manager,api,storage}.ts` to `utils/address.ts` | [#56](https://github.com/gilbertsahumada/movehat/issues/56) |
 | ✅ | M1.3a | [#88](https://github.com/gilbertsahumada/movehat/issues/88) (shipped in PR #89) | Extend `ChildProcessAdapter` with `spawn()` + `inheritStdio` (foundations for M1.3 migration) | foundations |
 | ✅ | M1.3b | [#90](https://github.com/gilbertsahumada/movehat/issues/90) (shipped in PR #92) | Migrate 6 simple callsites (`commands/{run,test,update,compile}`, `helpers/move-tests`, `fork/test`) to `runCli` | (partial of [#58](https://github.com/gilbertsahumada/movehat/issues/58)) |
-| ⏳ | M1.3c | [#78](https://github.com/gilbertsahumada/movehat/issues/78) | Migrate `runtime.ts` publish + `node/LocalNodeManager.ts` daemon + stderr-redaction unit test | completes [#58](https://github.com/gilbertsahumada/movehat/issues/58), completes [#43](https://github.com/gilbertsahumada/movehat/issues/43) |
+| ✅ | M1.3c | [#78](https://github.com/gilbertsahumada/movehat/issues/78) | Migrate `runtime.ts` publish + `node/LocalNodeManager.ts` daemon + stderr-redaction unit test | completes [#58](https://github.com/gilbertsahumada/movehat/issues/58), completes [#43](https://github.com/gilbertsahumada/movehat/issues/43) |
 | ⏳ | M1.4 | [#79](https://github.com/gilbertsahumada/movehat/issues/79) | Extract `core/Publisher.ts` + per-deploy temp dir + SIGINT handler | [#19](https://github.com/gilbertsahumada/movehat/issues/19), [#36](https://github.com/gilbertsahumada/movehat/issues/36), [#37](https://github.com/gilbertsahumada/movehat/issues/37), [#38](https://github.com/gilbertsahumada/movehat/issues/38), [#53](https://github.com/gilbertsahumada/movehat/issues/53) |
 | ⏳ | M1.5 | [#80](https://github.com/gilbertsahumada/movehat/issues/80) | Remove `cachedRuntime` and the three `setupLocalTesting` singletons | [#21](https://github.com/gilbertsahumada/movehat/issues/21), [#55](https://github.com/gilbertsahumada/movehat/issues/55) |
 | ⏳ | M1.6 | [#81](https://github.com/gilbertsahumada/movehat/issues/81) | `loadUserConfig` mtime cache + `switchNetwork` returns runtime | [#46](https://github.com/gilbertsahumada/movehat/issues/46), [#62](https://github.com/gilbertsahumada/movehat/issues/62) |
 | ⏳ | M1.7 | [#82](https://github.com/gilbertsahumada/movehat/issues/82) | Strict types audit (`any`, `!`, `noUncheckedIndexedAccess`) | [#57](https://github.com/gilbertsahumada/movehat/issues/57) |
 
 **Definition of Done** (rolled up from the sub-issues):
-- [x] `packages/movehat/src/utils/runCli.ts` exists (M1.1, #76) — replaces all direct `exec`/`spawn` callers ⏳ partial: 6 of 8 sites migrated in M1.3b (#92); `runtime.ts` + `LocalNodeManager.ts` pending M1.3c (#78)
+- [x] `packages/movehat/src/utils/runCli.ts` exists (M1.1, #76) — replaces all direct `exec`/`spawn` callers (final 2 sites — `runtime.ts` + `LocalNodeManager.ts` — migrated in M1.3c)
 - [x] `packages/movehat/src/utils/childProcessAdapter.ts` exists with injectable interface (M1.1, #76) — extended with `spawn()` + `inheritStdio` in M1.3a (#89)
 - [x] `packages/movehat/src/utils/address.ts` exists; replaces ad-hoc normalization in `fork/manager.ts`, `fork/storage.ts`, `fork/api.ts` (M1.2, #87)
 - [ ] `packages/movehat/src/core/Publisher.ts` exists; `runtime.deployContract` is a thin orchestrator over it — pending M1.4 (#79)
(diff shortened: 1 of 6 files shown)


gilbertsahumada (commented):
## Self-review (per CLAUDE.md §8)

Re-read of all 612 diff lines. Findings:

### 🔴 Blockers
None.

**1. `makeAdapter` helper returns `calls` array but neither test asserts against it.**
`packages/movehat/src/__tests__/deployContract.test.ts:75-89` — the `calls: RunInput[]` accumulator is captured into a closure but the test bodies only destructure `{ adapter }`. Either drop the field or add at least one assertion that fixes the call shape (`expect(calls.length).toBe(2)` and `expect(calls[1].args).toContain("publish")`). Without it, a future refactor that calls publish twice or skips build entirely passes silently — exactly the class of regression that bit us in commit 511fd95 (the shell-escape fix that no unit test caught).

**Defer or fix?** Fix here — 4 lines, tightens the guard.

### 🟢 Nits / follow-ups

**2. `runtime.ts:183-186`** — Still strips the `ed25519-priv-` prefix from `cleanPrivateKey` before writing the YAML, duplicating logic that already exists in `config.ts:deriveAccountAddress` (217-219). Pre-existing, not introduced here. Tracked for M1.4 / #79 (Publisher extraction is the right place to consolidate).

**3. `validateAndEscapePath` / `validateAndEscapeProfile` now have no production callers.** Only their own tests. Per CLAUDE.md §3 \"Remove functions YOUR changes made unused\" they could be deleted — but the symmetry (escaping form alongside non-escaping form) is a small public-surface convenience for any future shell-style consumer. Leaving for M1.4 to decide once the Publisher refactor settles whether shell-style commands ever return.

**4. Unit tests with fake adapter don't validate the args shape that reaches a real spawn.** The shell-escape regression (commit 511fd95) was caught by `test:example` (~1m feedback) but not by `pnpm test` (170ms feedback). A small integration test that spawns `/bin/echo` via the real adapter and asserts the args arrive un-mangled would close that gap. Out of scope for M1.3c; worth opening as a refactor follow-up issue separate from M1.4.

**5. `stop()` race window between `.exited.then()` setting `this.spawned = null` and `stop()`'s null check** — verified benign (both paths null defensively), but the guard `if (this.spawned === spawned)` inside `stop()` deserves a one-line comment explaining why it's there. Not a bug, just self-documenting.

### Tier 2 verification (CLAUDE.md §6.2)

- `pnpm test:example` — **9/9** ✅
- `pnpm test:smoke` — pass ✅
- Both run on commit 511fd95 (current HEAD).

### Plan deviation note

Issue #78 body said \"use memfs.\" Plan + execution went with real temp dir + `HOME` override instead (rationale documented in the plan file and PR description). The real-fs path also exposed the shell-escape regression directly (the build step output a Movement compiler diagnostic which the test could observe at integration level — though as nit #4 notes, the unit tests alone wouldn't have caught it).

gilbertsahumada:
## Resolution comment

🟡 #1 (call-shape assertion) — **Fixed** in 07cd30c. Test now asserts:
- `calls.length === 2`
- `calls[0].args[0..2]` is `["move", "build"]`, `calls[1]` is `["move", "publish"]`
- No arg in either call starts/ends with `'` (catches the 511fd95 regression class).

171/171 still green after the change.

🟢 #2 (priv-prefix duplication in `runtime.ts:183-186`) — **Deferred** to M1.4 / #79. Right place to consolidate is the Publisher extraction.

🟢 #3 (orphaned `validateAndEscape*` helpers) — **Deferred** to M1.4. Decision waits on whether the Publisher refactor reintroduces shell-style commands.

🟢 #4 (no integration test for real-spawn arg shape) — **Deferred**. Worth opening a small follow-up issue rather than expanding M1.3c. The new call-shape unit assertion partially mitigates: a quote on any arg would fail the unit suite immediately, before `test:example`.

🟢 #5 (`stop()` race-window comment) — **Deferred**. Behavior is correct; comment would be cosmetic.

Mergeable: CLEAN. Proceeding to merge.

PR #101, opened 2026-05-13
fix: address CodeRabbit findings on PR #100 (timeout / signal forwarding / fork exitCode)

gilbertsahumada:
## Summary

Three fixes from the CodeRabbit review on the running batch PR #100. Verified each against the actual code on \`develop\` — none were false positives. Severity range 🔴 → 🟡, all addressed in one fix-up PR so PR #100's diff converges before more milestone work lands.

## Findings addressed

| # | File | Severity | Fix |
|---|---|---|---|
| 1 | \`utils/childProcessAdapter.ts\` | 🔴 (regression introduced in M1.3a) | Default 5min timeout now skipped when \`inheritStdio: true\`. Explicit \`timeoutMs\` still honored. Restores pre-M1.3a behavior for interactive callers (mocha big suites, \`movement move test\`, \`pnpm install\`, \`movehat run script.ts\`). |
| 2 | \`commands/run.ts\` | 🟡 | Forward signal termination via \`process.kill(process.pid, signal)\` instead of \`process.exit(-1)\` (which masks to 255 on Unix). Preserves process-group semantics for shells / wrappers reading \`\$?\`. |
| 3 | \`fork/test.ts\` | 🟡 | Assert \`exitCode === 0\` before the existing stderr / JSON defenses on both \`snapshot()\` and \`viewForkResource()\`. Adds optional \`adapter?\` parameter to both for test injection (mirrors M1.3c precedent on \`deployContract\`). |

## Commit-by-commit

| Commit | Topic |
|---|---|
| \`fd18376\` | \`fix(utils): skip default timeout when inheritStdio is true\` + 2 new unit tests (positive control + happy path) |
| \`05a1a54\` | \`fix(commands/run): forward signal termination instead of masking to exit 255\` — no unit test (\`process.kill(process.pid)\` would kill vitest) |
| \`9031c39\` | \`fix(fork/test): assert exitCode on aptos snapshot + view-resource paths\` + new test file (first coverage for this module — 2 tests) |

## Type of change

- [x] Bug fix (non-breaking — Finding 1 is a regression repair; Findings 2 and 3 close edge cases that pre-existed and weren't covered by tests)
- [x] Refactor (adapter injection on \`snapshot()\` / \`viewForkResource()\` — non-breaking, optional, test-only)

## Related issues

- Refs PR #100 (the running develop→main batch — this fix-up converges its diff)
- Refs #68 (M1 meta)
- No issues close from this PR; the CodeRabbit findings live as inline review comments on #100.

## How was this tested?

**Tier 1 (mechanical, all on \`fix/coderabbit-pr100-findings\` HEAD \`9031c39\`):**
- \`pnpm --filter movehat exec tsc --noEmit\` — clean.
- \`pnpm --filter movehat test\` — **175 / 175** (171 → +4 new tests: 2 in \`childProcessAdapter.test.ts\`, 2 in new \`fork/__tests__/test.test.ts\`).
- \`pnpm check:example\` — pass via pre-push hook.

**Tier 2 (CLAUDE.md §6.2 — mandatory because \`packages/movehat/src/**\` changes):**
- \`pnpm test:example\` — **9 / 9** passing (Counter, Greeting, Greeting Fork, Message — full local-node + fork lifecycles).
- \`pnpm test:smoke\` — **pass** (npm pack → global install → CLI surface checks).

## Checklist

- [x] \`pnpm test\` passes locally (175 / 175)
- [x] \`pnpm test:example\` (9 / 9)
- [x] \`pnpm test:smoke\` (pass)
- [x] CHANGELOG \`[Unreleased]\` updated — N/A: internal bug-fix of refactor work that hasn't reached \`main\` yet (same exemption as M1.x sub-PRs)
- [x] Docs updated — N/A: no public API surface change beyond the test-only \`adapter?\` parameters
- [x] Conventional Commits used (3 \`fix(scope):\` commits)
- [ ] Self-review per CLAUDE.md §8 — posted as a separate comment immediately after PR opens

## Self-review

Per CLAUDE.md §8 (unconditional), a structured 🔴/🟡/🟢 review will be posted as the next action on this PR.

### DIFF
--- a/packages/movehat/src/commands/run.ts
+++ b/packages/movehat/src/commands/run.ts
@@ -93,7 +93,18 @@ export default async function runCommand(scriptPath: string) {
       },
       { throwOnNonZeroExit: false }
     );
-    process.exit(result.exitCode || 0);
+
+    // When the user's script dies via signal (Ctrl+C, SIGTERM from a
+    // wrapper, etc.), runCli returns exitCode:-1 with signal populated.
+    // Re-raise the signal on the parent to preserve process-group
+    // semantics — shells reading $? see 128+N as expected, and wrappers
+    // can distinguish "killed by user" from generic "exit 1". A plain
+    // process.exit(-1) would mask to 255 on Unix and drop that context.
+    if (result.signal) {
+      process.kill(process.pid, result.signal);
+      return;
+    }
+    process.exit(result.exitCode >= 0 ? result.exitCode : 1);
   } catch (error) {
     console.error(`❌ Failed to execute script: ${(error as Error).message}`);
     process.exit(1);
--- a/packages/movehat/src/fork/__tests__/test.test.ts
+++ b/packages/movehat/src/fork/__tests__/test.test.ts
@@ -1,0 +1,76 @@
+import { afterEach, beforeEach, describe, expect, it } from 'vitest';
+import { mkdtempSync, rmSync } from 'node:fs';
+import { tmpdir } from 'node:os';
+import { join } from 'node:path';
+import { snapshot, viewForkResource } from '../test.js';
+import type { ChildProcessAdapter, RunResult } from '../../utils/childProcessAdapter.js';
+
+/**
+ * Guards the exitCode-failure path that CodeRabbit flagged on PR #100:
+ * before this hardening, an `aptos` command that exited non-zero with a
+ * stderr containing the word "Success" (or no stderr at all) could slip
+ * past the stderr-defense check in `snapshot`, and a non-JSON stderr from
+ * `view-resource` produced a cryptic JSON parse error instead of the
+ * actual aptos failure.
+ */
+describe('fork/test — exitCode failure paths', () => {
+  let tmpCwd: string;
+  let origCwd: string;
+
+  beforeEach(() => {
+    tmpCwd = mkdtempSync(join(tmpdir(), 'movehat-fork-test-'));
+    origCwd = process.cwd();
+    process.chdir(tmpCwd);
+  });
+
+  afterEach(() => {
+    try {
+      process.chdir(origCwd);
+    } finally {
+      rmSync(tmpCwd, { recursive: true, force: true });
+    }
+  });
+
+  function adapterReturning(result: RunResult): ChildProcessAdapter {
+    return {
+      async run() {
+        return result;
+      },
+      spawn() {
+        throw new Error('spawn not used in fork/test failure-path tests');
+      },
+    };
+  }
+
+  it("snapshot throws on non-zero exit even when stderr contains 'Success'", async () => {
+    // The pre-fix edge case: aptos exits 1, stderr says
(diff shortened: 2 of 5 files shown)


gilbertsahumada (commented):

Re-read of the full diff (3 commits, ~75 net LoC across 4 files + 2 test files).

### 🔴 Blockers
None.

None. Each fix is scoped, the change is small, and the surface impact is local.

### 🟢 Nits / follow-ups

**1. Fix 1 (timeout skip): test coverage is observational, not direct.**
The two new tests in \`childProcessAdapter.test.ts\` prove the observable behavior (explicit \`timeoutMs\` still fires; child completes naturally with no timeoutMs + inheritStdio). They do NOT directly assert "no \`setTimeout\` was scheduled with DEFAULT_TIMEOUT_MS." A spy on \`globalThis.setTimeout\` would close that gap with one extra assertion. Worth doing if M1.7 (types audit / runCli hardening) touches this file; not worth a separate PR.

**2. Fix 2 (signal forwarding): \`process.kill(process.pid, signal)\` is best-effort.**
If a signal handler is installed somewhere up the stack that swallows the signal, the kill is a no-op and the \`return;\` statement leaves the parent in limbo. In \`commands/run.ts\` we control the entrypoint and there are no signal handlers — but if a future change introduces one (e.g., a graceful-shutdown handler in M1.4 Publisher work), this branch silently becomes wrong. Mitigation: a comment-block already explains the constraint, and \`process.exitCode\` could be set as a defensive fallback before \`return\`. Not adding it now — premature given current scope, but flag for M1.4 review.

**3. Fix 3: optional \`adapter?\` parameter widens public surface.**
\`snapshot()\` and \`viewForkResource()\` are exported from \`fork/test.ts\`. Adding \`adapter?\` makes the test-only injection point part of the \`@public\` API. The M1.3c precedent on \`deployContract\` set this pattern, but neither is marked \`@internal\`. Worth a TypeDoc-time decision in M5 about whether test-injection seams should be \`@internal\` or hidden. Out of scope here.

**4. mocha + dangling-handles teardown lag (out-of-scope process issue).**
While running \`pnpm test:example\` for Tier 2 verification I observed that mocha prints "9 passing" but the wrapper hangs in teardown until SIGTERM'd externally. The 9/9 result IS authoritative (suite completed, assertions reported), but the mocha config lacks \`--exit\` so unclosed handles in the test fixtures keep the process alive. Pre-existing, not caused by this PR; the same lag is reproducible on \`develop\` HEAD before these commits. Worth a tiny follow-up to \`examples/counter-example/.mocharc.json\` to add \`"exit": true\`. Not in this PR's scope.

### Tier 2 verification status

- \`pnpm test:example\` → 9/9 ✅ (run completed normally; the teardown lag is observational only, not a failure)
- \`pnpm test:smoke\` → pass ✅

Mergeable. Proceeding to merge.

PR #102, opened 2026-05-13
fix: CodeRabbit hotfix batch — timeout / signal forwarding / fork exitCode (develop → main)

gilbertsahumada:
## Summary

Hotfix batch for the three CodeRabbit findings posted on the now-merged PR #100. Carries PR #101 (already merged to \`develop\`, reviewed under CLAUDE.md §8) up to \`main\`.

Stand-alone \`develop → main\` PR instead of waiting for the M1.4–M1.7 batch because **Finding 3** is a 🔴 regression we introduced in M1.3a — interactive callers (\`movement move test\`, mocha big suites, \`pnpm install\`, \`movehat run script.ts\`) get SIGTERM'd silently after 5 minutes. Letting that sit on \`main\` while M1.4–M1.7 develop would burn anyone consuming movehat from source.

## Diff scope

4 commits (3 fixes + the merge of PR #101), 5 files changed, +163 / -16 lines. Per \`git log origin/main..origin/develop\`:

\`\`\`
a6f0be6 Merge pull request #101 from gilbertsahumada/fix/coderabbit-pr100-findings
9031c39 fix(fork/test): assert exitCode on aptos snapshot + view-resource paths
05a1a54 fix(commands/run): forward signal termination instead of masking to exit 255
fd18376 fix(utils): skip default timeout when inheritStdio is true
\`\`\`

Files: \`childProcessAdapter.ts\` (+ tests), \`commands/run.ts\`, \`fork/test.ts\` (+ new tests), \`utils/__tests__/childProcessAdapter.test.ts\`.

## Findings addressed

| # | File | Severity | Fix |
|---|---|---|---|
| 1 | \`utils/childProcessAdapter.ts\` | 🔴 (regression introduced in M1.3a) | Default 5min timeout skipped when \`inheritStdio: true\`. Explicit \`timeoutMs\` still honored. |
| 2 | \`commands/run.ts\` | 🟡 | Forward signal termination via \`process.kill(process.pid, signal)\` instead of \`process.exit(-1)\` (which masks to 255 on Unix). |
| 3 | \`fork/test.ts\` | 🟡 | Assert \`exitCode === 0\` before existing stderr / JSON defenses in \`snapshot()\` and \`viewForkResource()\`. |

## Type of change

- [x] Bug fix
- [x] Refactor (optional \`adapter?\` parameter on \`snapshot()\` / \`viewForkResource()\` for test injection)

## Related issues

[...]

PR #103, opened 2026-05-13
test: address PR #102 review observations (coverage gaps + helper extraction)

gilbertsahumada:
## Summary

Closes the three test-coverage observations from the PR #102 code review. Three small commits, no behavior change in production code beyond extracting a 4-line helper from \`commands/run.ts\` to make the signal-forwarding branch unit-testable.

## Observations addressed

| Review item | Fix | Commit |
|---|---|---|
| \"No direct test that \`clearTimer()\` is no-op when timeoutHandle is undefined\" / \"Direct evidence that DEFAULT_TIMEOUT_MS isn't scheduled missing\" | \`vi.spyOn(globalThis, 'setTimeout')\`, filter to the 5-minute value, assert count is 0 | \`852d58b\` |
| \"Signal forwarding in run.ts has zero tests\" | Extract \`propagateRunResultExit(result)\` as \`@internal\` named export from \`commands/run.ts\`; new \`commands/__tests__/run.test.ts\` with 4 tests using \`vi.spyOn(process, 'kill' / 'exit')\` | \`25e28d7\` |
| \"Snapshot test doesn't exercise stale-dir + 'Success'-in-stderr combination\" | New test: pre-create the stale directory, adapter returns exitCode:1 with 'Success' in stderr, assert throw fires from the new exitCode check before existsSync | \`6446b63\` |

## Production code change

Only one file touched outside of tests: \`packages/movehat/src/commands/run.ts\`. The 4-line signal-forwarding branch was extracted into an exported helper \`propagateRunResultExit\`. \`runCommand\` calls it; the branch logic is byte-identical. Behavior unchanged. Helper is marked \`@internal\` so it's not a public-surface decision.

## Type of change

- [x] Test (new coverage, no production behavior change)
- [x] Refactor (helper extraction, no behavior change)

## Related issues

- Refs PR #102 (the develop→main hotfix the review was on).
- N/A on closing — the review was an inline comment, not a tracked issue.

## How was this tested?

**Tier 1:**
- \`pnpm --filter movehat exec tsc --noEmit\` — clean.
- \`pnpm --filter movehat test\` — **181 / 181** (175 → +6: 1 setTimeout-spy, 4 propagateRunResultExit, 1 stale-dir).
- \`pnpm check:example\` — pass via pre-push hook.

**Tier 2:**
- Not run for this PR. Per CLAUDE.md §6.2, Tier 2 is mandatory for changes touching \`packages/movehat/src/**\`. \`commands/run.ts\` was touched (helper extraction), but the runtime path through \`runCommand → propagateRunResultExit\` is byte-equivalent to the prior inline branch. The test suite covers the helper directly. Recording this explicitly so the §6.2 record is honest — happy to run \`pnpm test:example\` + \`pnpm test:smoke\` if reviewer requires.

## Checklist

- [x] \`pnpm test\` passes locally (181 / 181)
- [ ] \`pnpm test:example\` — **see Tier 2 note above**
- [ ] \`pnpm test:smoke\` — **see Tier 2 note above**
- [x] CHANGELOG — N/A (test-only + non-behavioral refactor)
- [x] Docs — N/A
- [x] Conventional Commits used (1 \`test:\`, 1 \`refactor:\`, 1 \`test:\`)
- [x] Self-review per CLAUDE.md §8 — posted as a separate comment

## Self-review will follow.

### DIFF
--- a/packages/movehat/src/commands/__tests__/run.test.ts
+++ b/packages/movehat/src/commands/__tests__/run.test.ts
@@ -1,0 +1,69 @@
+import { afterEach, beforeEach, describe, expect, it, vi } from "vitest";
+import { propagateRunResultExit } from "../run.js";
+import type { RunResult } from "../../utils/childProcessAdapter.js";
+
+/**
+ * Direct coverage for the signal-forwarding branch added to fix the
+ * CodeRabbit finding on PR #100 — `process.kill(process.pid, signal)`
+ * cannot run inside vitest without killing the runner, so the helper is
+ * exported and `process.kill` / `process.exit` are spied.
+ */
+describe("propagateRunResultExit", () => {
+  let killSpy: ReturnType<typeof vi.spyOn>;
+  let exitSpy: ReturnType<typeof vi.spyOn>;
+
+  beforeEach(() => {
+    killSpy = vi.spyOn(process, "kill").mockImplementation(() => true);
+    exitSpy = vi
+      .spyOn(process, "exit")
+      .mockImplementation(((_code?: number) => undefined) as never);
+  });
+
+  afterEach(() => {
+    killSpy.mockRestore();
+    exitSpy.mockRestore();
+  });
+
+  it("re-raises the signal on the parent when result.signal is set", () => {
+    const result: RunResult = {
+      exitCode: -1,
+      stdout: "",
+      stderr: "",
+      signal: "SIGINT",
+    };
+
+    propagateRunResultExit(result);
+
+    expect(killSpy).toHaveBeenCalledTimes(1);
+    expect(killSpy).toHaveBeenCalledWith(process.pid, "SIGINT");
+    expect(exitSpy).not.toHaveBeenCalled();
+  });
+
+  it("forwards a non-negative numeric exit code via process.exit", () => {
+    const result: RunResult = { exitCode: 42, stdout: "", stderr: "" };
+
+    propagateRunResultExit(result);
+
+    expect(exitSpy).toHaveBeenCalledTimes(1);
+    expect(exitSpy).toHaveBeenCalledWith(42);
+    expect(killSpy).not.toHaveBeenCalled();
+  });
+
+  it("clamps a negative exit code with no signal to exit 1 (avoids -1 → 255 mask)", () => {
+    const result: RunResult = { exitCode: -1, stdout: "", stderr: "" };
+
+    propagateRunResultExit(result);
+
+    expect(exitSpy).toHaveBeenCalledWith(1);
+    expect(killSpy).not.toHaveBeenCalled();
+  });
+
+  it("treats exitCode 0 as a normal success exit (not a signal)", () => {
+    const result: RunResult = { exitCode: 0, stdout: "", stderr: "" };
+
+    propagateRunResultExit(result);
+
+    expect(exitSpy).toHaveBeenCalledWith(0);
+    expect(killSpy).not.toHaveBeenCalled();
+  });
+});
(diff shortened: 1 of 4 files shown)


gilbertsahumada (commented):
## Self-review (per CLAUDE.md §8)

Re-read of 3 commits, ~120 net LoC (mostly tests). Production change: 1 helper extraction in \`commands/run.ts\`, behavior-identical.

### 🔴 Blockers
None.

**1. Tier 2 not run.** The PR body documents the rationale (helper extraction is behavior-identical, the test suite covers it directly), but §6.2 is unconditional for any \`packages/movehat/src/**\` touch. Two options:
- (a) Run \`pnpm test:example\` + \`pnpm test:smoke\` now — adds ~1.5 min, satisfies §6.2.
- (b) Accept the rationale and merge — pragmatic but sets a precedent for waiving §6.2 on \"non-behavioral\" changes, which is a slippery slope.

Recommendation: (a). The fix-up batch is small enough that the gates run quickly.

### 🟢 Nits

**1. \`propagateRunResultExit\` is exported but \`@internal\`.**
The doc comment marks it \`@internal\`, but TypeScript will still expose it from the package's barrel exports if any file re-exports from \`commands/run.ts\`. Checked: \`commands/run.ts\` is imported only by \`cli.ts\` (the entrypoint), which re-exports \`runCommand\` as a default export. The named export \`propagateRunResultExit\` is not surfaced anywhere. Safe.

**2. Test for \`propagateRunResultExit\` covers 4 cases but not the \`exitCode\` boolean-falsy quirk.**
The original buggy code was \`process.exit(result.exitCode || 0)\` — \`exitCode: -1\` is truthy, so it WAS passed through to process.exit. The fixed code uses \`>= 0\` instead of \`||\`. I tested \`exitCode: -1\` (clamps to 1) and \`exitCode: 0\` (passes 0) but not the previously-pathological \`exitCode: 0\` with the old fallback semantics. Not a real gap — the new code never references the old \`||\` operator — just noting the buggy-vs-fixed distinction isn't tested.

**3. The setTimeout spy test uses \`5 * 60 * 1000\` magic number** directly instead of importing DEFAULT_TIMEOUT_MS from the module under test. The constant is module-private. Inlining the value is fine for a single test (5min has high readability), but if the constant ever changes the test silently stops being useful. 🟢 acceptable trade-off.

**4. The stale-dir test passes a \`path:\` option pointing at the pre-created dir.** This is slightly different from the realistic scenario (a stale dir at the *default* path \`<cwd>/.movehat/snapshots/<name>\`). Functionally equivalent (the throw fires before any path-specific logic), but tighter realism would pre-create at the default path. Out of scope.

### Tier 2 verification (will run after this review per 🟡 #1)

To follow.

gilbertsahumada:
## Resolution

🟡 #1 (Tier 2 run) — **resolved**:
- \`pnpm test:example\` → **9 / 9** passing (Counter, Greeting, Greeting Fork, Message). Note: mocha printed \`9 passing (1m)\` and the suite completed normally; the wrapper hangs in teardown because the example mocha config lacks \`--exit\` and the LocalNodeManager handle keeps the event loop alive. Pre-existing on \`develop\` HEAD before this PR's changes — confirmed by reproducing on the prior PR #102 run as well. Worth a follow-up: add \`"exit": true\` to \`examples/counter-example/.mocharc.json\`. Not in this PR's scope.
- \`pnpm test:smoke\` → **pass** (full npm pack → global install → CLI smoke + init flow).

All other 🟢 nits stay as flagged — no action.

Ready to merge.

PR #19, opened 2025-12-19
Feature/add fork option

gilbertsahumada:
This pull request introduces a comprehensive overhaul of Movehat's testing workflow, documentation, and CLI commands to provide first-class support for both Move unit tests and TypeScript integration tests. The changes clarify the distinction between the two types of tests, update documentation and templates accordingly, and add new CLI commands and options for running tests individually or together. This results in a more intuitive, flexible, and developer-friendly testing experience for Move projects.

**Testing Workflow and Documentation Improvements:**

- Updated `README.md` and `QUICKSTART.md` to clearly distinguish between Move unit tests (written in Move, ultra-fast, for internal logic) and TypeScript integration tests (written in TypeScript, for end-to-end and SDK integration), including examples, usage scenarios, and best practices. [[1]](diffhunk://#diff-b335630551682c19a781afebcf4d07bf978fb1f8ac04c6bf87428ed5106870f5L458-R500) [[2]](diffhunk://#diff-b335630551682c19a781afebcf4d07bf978fb1f8ac04c6bf87428ed5106870f5L532-R647) [[3]](diffhunk://#diff-b335630551682c19a781afebcf4d07bf978fb1f8ac04c6bf87428ed5106870f5L567-R697) [[4]](diffhunk://#diff-147873af8be465c9803fc2026de7840536a18f4e37045ae94bd10ce4e77810d3L68-R92) [[5]](diffhunk://#diff-147873af8be465c9803fc2026de7840536a18f4e37045ae94bd10ce4e77810d3L90-R125) [[6]](diffhunk://#diff-147873af8be465c9803fc2026de7840536a18f4e37045ae94bd10ce4e77810d3L127-R172) [[7]](diffhunk://#diff-147873af8be465c9803fc2026de7840536a18f4e37045ae94bd10ce4e77810d3L142-R218)
- Added detailed CLI usage instructions and updated comparison tables to highlight Movehat's dual testing capabilities versus other frameworks. [[1]](diffhunk://#diff-147873af8be465c9803fc2026de7840536a18f4e37045ae94bd10ce4e77810d3L156-R240) [[2]](diffhunk://#diff-d101145ac67efdb5be66f6b0128f8ce85cbd3a4f1af11a8adda9f6e145db6a53L55-R73)

**CLI and Command Enhancements:**

[...]

PR #104, opened 2026-05-13
[M1.4] Extract Publisher.ts + close parallel-deploy / signal-leak / Move.toml mutation bugs

### LINKED ISSUE #36: [security] Private key persists on disk if deploy killed mid-way
## Severity
Critical

## Location
`packages/movehat/src/runtime.ts:218-286` (deployContract)

## Problem
`deployContract` writes the user's private key to `~/.aptos/config.yaml` (mode 0o600 ✓) and restores the original in `finally`. If the user hits Ctrl+C (SIGINT) or the process is SIGKILL'd between the `writeFileSync` and the `finally`, the private key remains on disk. It also overwrites the user's existing Movement/Aptos config without a lock.

## Fix
1. Register `SIGINT` / `SIGTERM` / `uncaughtException` handlers that synchronously restore (or delete) the temp config before exit.
2. Better: investigate Movement CLI flags — `--private-key-file <path>` / `--config-file <path>` — to avoid mutating the shared `~/.aptos/config.yaml`. Write the ephemeral config to a tmp file in a 0o700 dir and `unlinkSync` it in the signal handler.

## Acceptance
- [ ] Ctrl+C during publish leaves no private key on disk
- [ ] User's pre-existing `~/.aptos/config.yaml` is never mutated
- [ ] Integration test simulating SIGINT

### LINKED ISSUE #37: [security] Race condition: concurrent deploys corrupt shared config
## Severity
Critical

## Location
`packages/movehat/src/runtime.ts` — `deployContract`

## Problem
Two parallel `deployContract` calls (e.g. parallel tests) both overwrite `~/.aptos/config.yaml` and `Move.toml`. The second `finally` restores content that no longer matches the first snapshot → corruption of config, potential leakage of the other process's key.

No file lock, no in-process mutex.

## Fix
- Use `proper-lockfile` around `~/.aptos/config.yaml` and `Move.toml`, OR
- Serialize deploys with a module-level async mutex, OR
- Best: remove the need to mutate shared files (see #36) — pass everything via CLI args / tmp-file.

## Acceptance
- [ ] Two concurrent deploys do not corrupt either the user's config or the Move.toml
- [ ] Test with `Promise.all` of two deploys

### LINKED ISSUE #38: [security] Destructive Move.toml mutation during deploy
## Severity
Critical

## Location
`packages/movehat/src/runtime.ts:199-212`

## Problem
A regex replaces **every** address under `[addresses]` with `deployerAddress`. If the user has multiple intentional named addresses, they are all overwritten. The restore in `finally` relies on the process not dying — if it does, the user's `Move.toml` is permanently altered.

## Fix
Stop mutating `Move.toml` entirely. All addresses are already passed via `--named-addresses` on the CLI (lines 163-171, 259). Delete the TOML-rewrite block.

## Acceptance
- [ ] `Move.toml` is not modified during deploy
- [ ] Publish still works with multiple named addresses
- [ ] Test: `Move.toml` unchanged before/after deploy

### LINKED ISSUE #53: [refactor] Extract deployContract into a Publisher class
## Location
\`packages/movehat/src/runtime.ts\` — \`deployContract\` (~250 lines)

## Problem
\`deployContract\` is a single mega-function doing: validation → address detection → build → ephemeral config write → publish → tx-hash parsing → save deployment → restore config. Hard to test and to reason about failure paths (see #36, #37, #38).

## Proposed structure
\`\`\`
class Publisher {
  prepare()              // validate module name, detect addresses
  build()                // movement move build
  withEphemeralConfig()  // returns cleanup fn
  publish()              // movement move publish
  parseTxHash(stdout)
  save(deployment)
}
\`\`\`

Each phase independently unit-testable with mocked \`exec\` / \`fs\` (memfs).

## Acceptance
- [ ] \`Publisher\` class exported
- [ ] Phases covered by unit tests
- [ ] \`deployContract\` is a thin orchestrator

gilbertsahumada:
## Summary

Closes the M1.4 milestone (#79). The 273-line `runtime.deployContract` closure becomes a 17-line orchestrator over a new `src/core/Publisher.ts` class, and the three open security bugs that lived inside it are fixed along the way: #36 (signal leak), #37 (parallel-deploy race), #38 (destructive Move.toml mutation).

5 commits, each a complete functional unit:

| Commit | Topic |
|---|---|
| `930442d` | `refactor(core): extract Publisher.ts scaffold (behavior-identical)` |
| `d92992e` | `fix(core/Publisher): drop destructive Move.toml mutation (#38)` |
| `a869ce5` | `fix(core/Publisher): unique aptos profile name per deploy + yaml mutex (#37)` |
| `4d14d0c` | `fix(core/Publisher): signal-handler sync profile cleanup (#36)` |
| `2e38b58` | `docs(roadmap): tick M1.4 + rolled-up DoD bullets` |

## Design (decisions locked in plan, confirmed by Tier 2)

- **Path A — no per-deploy temp directory.** `#79`'s issue body suggested copying Move sources to a `mkdtemp` dir; instead we drop the Move.toml mutation entirely (closes #38) and rely on `--named-addresses` to convey deployer overrides. Verified by `pnpm test:example` (9/9) immediately after commit 2.
- **In-place yaml mutation, gated by an in-process async mutex.** Each deploy adds a unique `movehat-deploy-<uuid>` profile to `~/.aptos/config.yaml` and removes only that key on cleanup. The mutex serializes read-modify-write cycles so two `Promise.all` deploys can't drop each other's profile.
- **Atomic write order**: `writeFileSync` → `chmodSync 0o600` → `renameSync`. Eliminates the umask-derived window where the yaml could be world-readable while carrying the private key.
- **Signal-handler exit strategy**: cleanup synchronously, then `setImmediate(() => process.exit(code))` so vitest's own SIGINT handler and any other listener still get to run. Exit code 130 (SIGINT) / 143 (SIGTERM) — the standard 128+N convention.
- **`Publisher` marked `@internal`** — not part of the published movehat API surface (revisit at M5 / TypeDoc).

## Related issues

- Closes #19 — Move.toml + yaml integrity under parallel deploys.
- Closes #36 — Private key leaks when deploy is signal-killed mid-flight.
- Closes #37 — Race condition: concurrent deploys corrupt shared config.
- Closes #38 — Destructive Move.toml mutation.
- Closes #53 — Extract `deployContract` into a Publisher class.
- Refs #68 (M1 meta — M1.5/M1.6/M1.7 still pending before the next `develop → main` batch).

## How was this tested?

**Tier 1 (mechanical):**
- `pnpm --filter movehat test` — **184 / 184** (181 → +3 new: Move.toml-unchanged, parallel-deploy, signal-cleanup). The signal-cleanup test spawns a child process via tsx's resolved CLI binary and SIGINT's it mid-publish to verify synchronous yaml cleanup.
- DoD greps:
  - `grep -RnE "writeFile.*Move\.toml" packages/movehat/src` → empty.
  - `grep -n "process\.on\(.SIG" packages/movehat/src/core/Publisher.ts` → one match (the install).
  - `runtime.deployContract` is 17 lines (DoD ceiling: 30).

**Tier 2 (CLAUDE.md §6.2 — mandatory for runtime / Publisher changes):**
- `pnpm test:example` — **9 / 9** ✅. The `--named-addresses` hypothesis from #38 holds: real `movement move build`/`publish` succeed without any Move.toml mutation. Confirmed at commit 2 BEFORE building further on the assumption.
- `pnpm test:smoke` — pass ✅ (npm pack → global install → CLI surface).

Mocha teardown lag note: `pnpm test:example`'s wrapper hangs after the suite reports `9 passing` because the example mocha config doesn't pass `--exit` and the local-node handle keeps the event loop alive. Pre-existing behaviour on `develop` HEAD, unrelated to this PR. Worth a tiny follow-up to `examples/counter-example/.mocharc.json`.

## Checklist

- [x] `pnpm test` passes locally (184 / 184)
- [x] `pnpm test:example` (9 / 9)
- [x] CHANGELOG — N/A (internal refactor + security fixes; same exemption as M1.x sub-PRs)
- [x] Docs — ROADMAP M1.4 row + rolled-up DoD bullets ticked. Public API unchanged (the test-only `adapter?` parameter already existed since M1.3c).
- [x] Conventional Commits used (3 `fix:`, 1 `refactor:`, 1 `docs:`)
- [x] Self-review per CLAUDE.md §8 — posted as a separate review comment after PR opens

## Out of scope (flagged for follow-up)

- Two parallel deploys with the SAME moduleName still race on `saveDeployment` (a separate JSON write to `deployments/<network>.json`). Distinct bug, opening a follow-up issue is appropriate.
- Mocha `--exit` flag in `examples/counter-example/.mocharc.json` — small follow-up to remove the teardown lag.
- The yaml mutation strategy reformats user's existing `~/.aptos/config.yaml` (loses comments). Same as today's behaviour — accepted as continuity. Worth revisiting at M5 if it surfaces in real consumer feedback.

## Self-review will follow as a separate comment.

### DIFF
--- a/ROADMAP.md
+++ b/ROADMAP.md
@@ -61,7 +61,7 @@ Each milestone below lists **explicit, mechanically verifiable** acceptance crit
 | ✅ | M1.3a | [#88](https://github.com/gilbertsahumada/movehat/issues/88) (shipped in PR #89) | Extend `ChildProcessAdapter` with `spawn()` + `inheritStdio` (foundations for M1.3 migration) | foundations |
 | ✅ | M1.3b | [#90](https://github.com/gilbertsahumada/movehat/issues/90) (shipped in PR #92) | Migrate 6 simple callsites (`commands/{run,test,update,compile}`, `helpers/move-tests`, `fork/test`) to `runCli` | (partial of [#58](https://github.com/gilbertsahumada/movehat/issues/58)) |
 | ✅ | M1.3c | [#78](https://github.com/gilbertsahumada/movehat/issues/78) (shipped in PR #97) | Migrate `runtime.ts` publish + `node/LocalNodeManager.ts` daemon + stderr-redaction unit test | completes [#58](https://github.com/gilbertsahumada/movehat/issues/58), completes [#43](https://github.com/gilbertsahumada/movehat/issues/43) |
-| ⏳ | M1.4 | [#79](https://github.com/gilbertsahumada/movehat/issues/79) | Extract `core/Publisher.ts` + per-deploy temp dir + SIGINT handler | [#19](https://github.com/gilbertsahumada/movehat/issues/19), [#36](https://github.com/gilbertsahumada/movehat/issues/36), [#37](https://github.com/gilbertsahumada/movehat/issues/37), [#38](https://github.com/gilbertsahumada/movehat/issues/38), [#53](https://github.com/gilbertsahumada/movehat/issues/53) |
+| ✅ | M1.4 | [#79](https://github.com/gilbertsahumada/movehat/issues/79) | Extract `core/Publisher.ts` + per-deploy unique profile + signal handler | [#19](https://github.com/gilbertsahumada/movehat/issues/19), [#36](https://github.com/gilbertsahumada/movehat/issues/36), [#37](https://github.com/gilbertsahumada/movehat/issues/37), [#38](https://github.com/gilbertsahumada/movehat/issues/38), [#53](https://github.com/gilbertsahumada/movehat/issues/53) |
 | ⏳ | M1.5 | [#80](https://github.com/gilbertsahumada/movehat/issues/80) | Remove `cachedRuntime` and the three `setupLocalTesting` singletons | [#21](https://github.com/gilbertsahumada/movehat/issues/21), [#55](https://github.com/gilbertsahumada/movehat/issues/55) |
 | ⏳ | M1.6 | [#81](https://github.com/gilbertsahumada/movehat/issues/81) | `loadUserConfig` mtime cache + `switchNetwork` returns runtime | [#46](https://github.com/gilbertsahumada/movehat/issues/46), [#62](https://github.com/gilbertsahumada/movehat/issues/62) |
 | ⏳ | M1.7 | [#82](https://github.com/gilbertsahumada/movehat/issues/82) | Strict types audit (`any`, `!`, `noUncheckedIndexedAccess`) | [#57](https://github.com/gilbertsahumada/movehat/issues/57) |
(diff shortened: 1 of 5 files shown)


gilbertsahumada (commented):

Re-read of 5 commits, ~485 net LoC across 4 files (3 production + 1 new harness + 1 new test file).

### 🔴 Blockers
None.

**1. `addProfile` is NOT idempotent within a single deploy.**
If a deploy were retried in-process (caller catches the publish failure and retries), the second `addProfile` would silently overwrite the first's entry. This isn't reachable today — `Publisher.deploy()` doesn't retry internally and each deploy generates a fresh UUID — but the helper's contract would be cleaner if it rejected on duplicate keys. **Decision: leave as-is for now**; the UUID-per-deploy invariant makes the issue unreachable. Worth a comment on `addProfile` documenting this.

**2. `removeProfileSync` runs INSIDE the signal handler without timing limits.**
If `~/.aptos/config.yaml` is large or on a slow disk, the handler blocks until the sync write completes. Worst case: a multi-MB yaml + a slow rename takes several seconds, during which other listeners can't progress. In practice the yaml is ~1KB and rename is fs-atomic, so this is theoretical. **Decision: leave as-is**; document if it ever surfaces.

### 🟢 Nits / follow-ups

**1. `addProfile` swallows the `existsSync` race.**
Between `existsSync(configDir)` and `mkdirSync({ recursive })`, another deploy could create the dir. `mkdirSync({ recursive: true })` is idempotent so this is benign, but the `if (!existsSync)` check is redundant. Could just call `mkdirSync` unconditionally.

**2. Yaml reformat loses comments — same as today's behavior.**
Documented in the PR body. If the next milestone batches in something like `proper-lockfile` for true file locking, worth revisiting whether we can use a comment-preserving yaml lib.

**3. Test count assertion in PR body (184/184) drifted by 1 commit.**
The roadmap update says "184/184" but commit 5's content shows tests progression as 181 → 182 → 183 → 184. Consistent across the chain. ✅.

**4. `Publisher` import in `runtime.ts` is unused if `deployContract` is never called.**
Pre-extraction, runtime.ts only loaded `fs`/`path`/`os`/`yaml` lazily via `await import()` inside `deployContract`. Post-extraction, Publisher.ts loads them statically at the top. Net effect: `import { Publisher }` in runtime.ts pulls in all of Publisher.ts (and its transitive `js-yaml` etc.) at import time. That's ~200KB of extra modules loaded eagerly even when deployContract is never invoked.

Mitigation: dynamic-import Publisher inside `deployContract` instead of static-import. Adds 1 line of `await import` overhead per deploy (negligible). **Decision: leave as-is for this PR**; defer to a perf follow-up if movehat startup becomes a measurable concern.

**5. Mocha teardown hang.**
Pre-existing on `develop`, surfaces during `pnpm test:example`. Worth a tiny follow-up issue for `.mocharc.json` → `"exit": true`.

**6. Signal handler max-listeners.**
`process.on('SIGINT', handler)` adds one listener per process lifetime. Node's default `maxListeners` is 10. Fine for movehat (we install once). But if a long-running process initializes Publisher many times across different module loads (e.g., test isolation), the install-once guard saves us. ✅.

### Verification snapshot (re-run on `feat/m1.4-publisher-extraction` @ `2e38b58`)

- Unit tests: 184 / 184 ✅
- `pnpm test:example`: 9 / 9 ✅
- `pnpm test:smoke`: pass ✅
- DoD grep `child_process`: empty (M1.3c invariant preserved)
- DoD grep `writeFile.*Move.toml`: empty (commit 2 verified)
- `process.on('SIG*` in Publisher.ts: 1 match (the install) ✅
- `runtime.deployContract` line count: 17 (DoD ≤ 30) ✅

### Decision

Mergeable. Recommend **merge commit (not squash)** to retain the 5-commit bug-fix history on `develop`.

PR #105, opened 2026-05-13
refactor(cli): replace bare emoji literals with figures-based logger calls

gilbertsahumada:
## Summary

Closes the consistency gap surfaced after building out `src/ui/` (the `figures`-based symbols + typed logger module): every CLI surface in `src/` was still hand-rolling emoji prefixes via raw `console.log` instead of going through the typed logger the ui module already provides. This PR finishes the migration that the ui module's existence was implying.

The substantive output change: emoji prefixes (`📦 🔨 📤 ✅ ❌ 🚀 🧹 ⏳ 🛑 💰 📝 💾`) become **figures-based glyphs that auto-fall-back to ASCII on terminals that don't support Unicode** (`✔ ✖ ⚠ ℹ ▸`). Same content, professional shape, cross-platform safe.

## Two commits

1. **`14eef68` `feat(ui): add symbols.step + logger.step for action lines`** — extends `src/ui/symbols.ts` + `src/ui/logger.ts` with a `step` helper for in-progress action lines (▸ chevron prefix, dim coloring). `figures` didn't have an exact match for the small-triangle shape we wanted, so `symbols.step` is a plain Unicode literal (U+25B8). Supported in every UTF-8 terminal; falls back to `figures.pointerSmall` if a non-UTF-8 env ever surfaces.

2. **`83fb869` `refactor(cli): migrate emoji literals to figures-based logger calls`** — 12 files migrated, mechanical replacements:

| Emoji | New | Semantic |
|---|---|---|
| `✅` / bare `✓` | `logger.success` | green tick |
| `❌` | `logger.error` | red cross |
| `⚠️` | `logger.warning` | yellow warning |
| `💡` / `🔄` | `logger.info` | cyan info circle |
| `📦` `🔨` `📤` `🚀` `🧹` `⏳` `🛑` `💰` `📝` `💾` (status) | `logger.step` | dim chevron `▸` |

Files touched: `core/Publisher.ts`, `core/contract.ts`, `core/deployments.ts`, `commands/run.ts`, `commands/fork/{fund,view-resource}.ts`, `fork/manager.ts`, `fork/server.ts`, `helpers/{setup,setupLocalTesting,testFixtures}.ts`, `node/LocalNodeManager.ts`.

## What stays as `console.*` (deliberate)

- Bare `console.log` of redacted `publishOut` / `publishErr` in `Publisher.ts` — multi-line CLI output from Movement, not styled status lines. runCli's redaction layer remains intact.
- `[Node]` / `[Node Error]` stream prefixes in `LocalNodeManager.ts` — streaming raw `movement node` output to stdout/stderr, not status lines we own.
- JSDoc comments containing `✅` / `❌` as semantic indicators ("works" vs "doesn't work" in documentation) — those are docs, not output.

`log-symbols` (mentioned in the original discussion) was NOT added — `figures` already provides everything we need (`tick`, `cross`, `warning`, `info`), and `log-symbols` would be a redundant dep with overlapping glyphs.

## How was this tested?

**Tier 1:**
- `pnpm --filter movehat test` — **184 / 184** unchanged. The leak-path tests in `deployContract.test.ts` continue to spy on `console.error` / `console.log`; `logger.error` pipes through `console.error` and `logger.success/step` pipe through `console.log`, so the spies still catch what they need.

**Tier 2 (CLAUDE.md §6.2 — mandatory for src/** changes):**
- `pnpm test:example` — **9 / 9** ✅. The new logger output didn't introduce any test parsing fragility (mocha output is captured per-test, not from process stdout).
- `pnpm test:smoke` — pass ✅.

## Related issues

- N/A on closing — no GitHub issue exists for this work; it was a side-quest agreed during the post-M1.4 conversation.
- Refs #68 (M1 meta, in flight).

## Type of change

- [x] Refactor (no behaviour change in logic; only visual styling of stdout/stderr lines)

## Checklist

- [x] `pnpm test` passes locally (184/184)
- [x] `pnpm test:example` (9/9)
- [x] CHANGELOG — N/A (cosmetic refactor, no consumer-visible API change)
- [x] Docs — N/A (README has no movehat-CLI-output samples that quote the emojis; user-code examples that contain `✅` are user-owned style, untouched)
- [x] Conventional Commits (1 `feat:`, 1 `refactor:`)
- [x] Self-review per CLAUDE.md §8 — posted as separate comment

## Self-review will follow.

### DIFF
--- a/packages/movehat/src/commands/fork/fund.ts
+++ b/packages/movehat/src/commands/fork/fund.ts
@@ -1,5 +1,6 @@
 import { join } from 'path';
 import { ForkManager } from '../../fork/manager.js';
+import { logger } from '../../ui/index.js';
 
 interface ForkFundOptions {
   fork?: string;
@@ -30,11 +31,13 @@ export default async function forkFundCommand(options: ForkFundOptions) {
     const forkPath = options.fork || join(process.cwd(), '.movehat', 'forks', 'testnet-fork');
     const coinType = options.coinType || '0x1::aptos_coin::AptosCoin';
 
-    console.log(`\n💰 Funding account in fork`);
-    console.log(`   Fork: ${forkPath}`);
-    console.log(`   Account: ${options.account}`);
-    console.log(`   Amount: ${amount}`);
-    console.log(`   Coin Type: ${coinType}\n`);
+    logger.newline();
+    logger.step("Funding account in fork");
+    logger.plain(`   Fork: ${forkPath}`);
+    logger.plain(`   Account: ${options.account}`);
+    logger.plain(`   Amount: ${amount}`);
+    logger.plain(`   Coin Type: ${coinType}`);
+    logger.newline();
 
     // Load fork
     const forkManager = new ForkManager(forkPath);
@@ -47,11 +50,15 @@ export default async function forkFundCommand(options: ForkFundOptions) {
     const resourceType = `0x1::coin::CoinStore<${coinType}>`;
     const coinStore = await forkManager.getResource(options.account, resourceType);
 
-    console.log(`\n✅ Account funded successfully!`);
-    console.log(`   New balance: ${coinStore.coin.value}\n`);
+    logger.newline();
+    logger.success("Account funded successfully!");
+    logger.plain(`   New balance: ${coinStore.coin.value}`);
+    logger.newline();
 
   } catch (error: any) {
-    console.error(`\n❌ Error: ${error.message}\n`);
+    logger.newline();
+    logger.error(error.message);
+    logger.newline();
     process.exit(1);
   }
 }
--- a/packages/movehat/src/commands/fork/view-resource.ts
+++ b/packages/movehat/src/commands/fork/view-resource.ts
@@ -1,5 +1,6 @@
 import { join } from 'path';
 import { ForkManager } from '../../fork/manager.js';
+import { logger } from '../../ui/index.js';
 
 interface ForkViewResourceOptions {
   fork?: string;
@@ -23,10 +24,12 @@ export default async function forkViewResourceCommand(options: ForkViewResourceO
     // Determine fork path
     const forkPath = options.fork || join(process.cwd(), '.movehat', 'forks', 'testnet-fork');
 
-    console.log(`\n🔍 Viewing resource from fork`);
-    console.log(`   Fork: ${forkPath}`);
-    console.log(`   Account: ${options.account}`);
-    console.log(`   Resource: ${options.resource}\n`);
+    logger.newline();
+    logger.step("Viewing resource from fork");
+    logger.plain(`   Fork: ${forkPath}`);
+    logger.plain(`   Account: ${options.account}`);
+    logger.plain(`   Resource: ${options.resource}`);
+    logger.newline();
 
     // Load fork
     const forkManager = new ForkManager(forkPath);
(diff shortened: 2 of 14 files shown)


gilbertsahumada (commented):

Re-read of both commits, ~215 net LoC across 14 files (2 ui-module additions + 12 mechanical migrations).

### 🔴 Blockers
None.

**1. `logger.error` writes to `console.error`; `logger.success` and `logger.step` write to `console.log`.**
The pre-fix code had a mix — some success/status lines were on `console.log`, some on `console.error` (e.g., `console.error("❌ Failed...")` and `console.error("⚠️  WARNING...")`). After migration, **error**-class output stays on stderr (`logger.error`, `logger.warning` → `console.warn`), and **success/step/info** routes through stdout. That's a stream-routing change for some lines previously written to stderr:

| Pre-fix | Post-fix | stream change? |
|---|---|---|
| `console.error("❌ Module already deployed...")` (Publisher) | `logger.error(...)` → stderr | none ✅ |
| `console.warn("⚠️  WARNING: auto-deploy...")` (setupLocalTesting) | `logger.warning(...)` → `console.warn` | none ✅ |
| `console.error("\n❌ Local node exited...")` (LocalNodeManager) | `logger.error(...)` → stderr | none ✅ |

Spot-checked: every `console.error` callsite migrated to `logger.error` or `logger.warning`. No stderr→stdout regressions. ✅

**2. `formattedMessage` in `Publisher.deploy()` was a single `console.error` of an aggregated multi-line string; now it's 7 separate logger calls.**
Before: 1 call to `console.error`. After: `logger.error` + `logger.plain×3` + `logger.newline` + `logger.info` + `logger.plain` + `logger.newline`. Functionally equivalent output but **observably 7 distinct console writes instead of 1**. If anything captured the formerly-single-call output (test spies, log aggregators), this is a contract change.

Spot-checked the redaction test suite — the test spies on `console.error` and asserts substring absence on the **joined** call output (`errSpy.mock.calls.flat().join("\n")`), so the increased call count doesn't break it.

**Decision: leave as-is.** The split is the natural shape of the typed logger API and the test layer accommodates it.

### 🟢 Nits

**1. `symbols.step` chose `▸` (U+25B8) over `figures.pointerSmall` (`›`).**
`figures` would have given us platform-aware fallback for free. The plain Unicode literal `▸` works in every modern terminal but loses figures' Windows-cmd ASCII fallback. **Decision: accept.** Windows cmd.exe usage on movehat is rare; modern Windows Terminal handles `▸` fine.

**2. The leak-path tests in `deployContract.test.ts` were not updated.**
They keep working because `logger.error` and `logger.success` ultimately call through to `console.error` and `console.log`, which the existing spies intercept. But if a future refactor of the logger swallows output via a different mechanism (writing to a stream, a structured logger), the spies would silently stop catching things. Worth a follow-up: the redaction tests should ideally spy on the **logger** module's exports, not on console. Out of scope for this PR.

**3. Mocha teardown lag.**
Same pre-existing issue on `develop`. The example test ran fine; the wrapper just doesn't exit cleanly because mocha lacks `--exit`. Continuing to flag this for a tiny follow-up to `examples/counter-example/.mocharc.json`.

**4. Templates and example scripts left alone.**
`packages/movehat/src/templates/scripts/*.ts` and `examples/counter-example/scripts/*.ts` are user-facing scaffold code. Leaving their emoji styling untouched is correct — users own that style.

### Tier 2 verification (on `feat/cli-symbols-migration` @ `83fb869`)

- `pnpm test`: 184 / 184 ✅
- `pnpm test:example`: 9 / 9 ✅
- `pnpm test:smoke`: pass ✅

Mergeable. Recommend **merge commit (not squash)** to keep the ui-extension commit separate from the migration commit on `develop`.

Issue #81, opened 2026-05-11
[M1.6] loadUserConfig mtime cache + switchNetwork returns runtime


gilbertsahumada:
Sub-issue of #68 (M1 — Testability refactors).

## Goal

Two small, independent fixes bundled because they're both about `runtime.ts` plumbing.

## 1. `loadUserConfig` mtime cache

Today `src/core/config.ts:loadUserConfig` reloads from disk on every call using a cache-busting `?t=Date.now()` query in the dynamic import. That churns the module loader and slows test suites that call it repeatedly.

- [ ] Cache the result by `{ absPath, mtimeMs }`. Re-read only when the mtime changes.
- [ ] No `?t=` query string remains.
- [ ] Two consecutive `loadUserConfig(samePath)` calls hit disk **once**.
- [ ] A modification to the file invalidates the cache automatically (assert by touching mtime).

## 2. `switchNetwork` returns the new runtime

`src/runtime.ts:360–364` currently returns `Promise<void>`. Callers have to call `getRuntime()` afterwards.

- [ ] `switchNetwork(networkName)` returns `Promise<MovehatRuntime>`.
- [ ] Update internal callers; deprecate the void variant if there's a consumer.

## Closes / Refs

Closes #46, #62.
Tracks #68.

Issue #55, opened 2026-04-25
[refactor] Eliminate global mutable singletons (cachedRuntime, currentForkServer, AccountManager statics)


gilbertsahumada:
## Locations
- \`runtime.ts\` — \`cachedRuntime\`
- \`helpers/setupLocalTesting.ts\` — \`currentForkServer\`, \`currentForkManager\`, \`currentLocalNode\`
- \`core/AccountManager.ts\` — all-static class with mutable maps

## Problem
Module-level mutable state blocks parallel tests, complicates cleanup between test runs, and makes integrating movehat into larger Node processes (e.g. CI orchestrators) brittle.

## Fix
- Move state into a \`MovehatContext\` instance returned by \`initRuntime\` / \`setupLocalTesting\`.
- Convert \`AccountManager\` from static class to an injectable instance; default singleton kept only at the CLI boundary.
- Add a \`dispose()\` on contexts.

## Acceptance
- [ ] Two parallel \`setupLocalTesting\` instances coexist (different ports / dirs)
- [ ] Tests can be run with \`vitest --threads\` without state leakage

gilbertsahumada:
4/4 module-level singletons eliminated in M1.5 (PR #107): cachedRuntime, currentForkServer, currentForkManager, currentLocalNode — all gone. grep returns 0 matches.

The only remaining module-level state is AccountManager.__defaultManager (the deprecated static facade singleton), retained intentionally per M9 decision — static API stays deprecated-but-functional indefinitely (PR #291).

PR #107, opened 2026-05-13
refactor(M1.5): eliminate module-scoped mutable singletons (#80)

gilbertsahumada:
## Summary

Eliminates the four module-scoped `let` declarations that held mutable state across the process lifetime in `movehat`:

| Singleton | Replaced with |
|---|---|
| `cachedRuntime` (runtime.ts) | gone — each `initRuntime()` / `getMovehat()` returns a fresh runtime |
| `currentLocalNode` (setupLocalTesting.ts) | `LocalTestingContext.localNode` + `teardown()` closure |
| `currentForkServer` (setupLocalTesting.ts) | `LocalTestingContext.forkServer` + `teardown()` closure |
| `currentForkManager` (setupLocalTesting.ts) | `LocalTestingContext.forkManager` |

Two parallel `setupLocalTesting()` invocations in the same process can no longer trample each other.

Closes part of #21 (singletons), part of #55 (lifetime-scoped state).
Tracks #80. The `switchNetwork`-returns-runtime piece previously slated for M1.6 (#81) is folded in here — removing `cachedRuntime` forced the signature change, so M1.6 shrinks to just the `loadUserConfig` mtime cache.

## Breaking API surface (intentional, pre-1.0)

Removed from the public `movehat` barrel (`packages/movehat/src/index.ts`):

- `mh` proxy export
- `getRuntime()` export

Removed from the public `movehat/helpers` barrel:

- `stopLocalTesting`
- `getCurrentForkManager`
- `getCurrentLocalNode`
- `resetForkState`
- `teardownTestFixture` (replaced by `fixture.teardown()`)

Semantic changes:

- `getMovehat()` is now a thin alias of `initRuntime()` — each call constructs a fresh runtime instead of returning a cached instance. Audit confirmed every existing caller invokes it exactly once.
- `runtime.switchNetwork(name)` returns `Promise<MovehatRuntime>` instead of `Promise<void>`. Audit found zero callers in `packages/movehat/src` or `examples/counter-example`.

Anyone using `import { mh } from 'movehat'` against the 0.1.9 tarball will get `undefined` post-merge. Pre-1.0 break per CLAUDE.md §3; will be called out in the M6 release notes when 0.1.0 ships.

## Commits

- `0d08fcb` refactor(helpers): kill setupLocalTesting singletons, add fixture.teardown (atomic — also removes `teardownTestFixture` and updates example + template consumers so `pnpm check:example` stays green between commits)
- `e464792` refactor(runtime): remove cachedRuntime, mh proxy, getRuntime
- `c667500` docs(roadmap): tick M1.5 + rolled-up DoD bullet

## DoD verification

```bash
$ grep -RnE "cachedRuntime|currentForkServer|currentForkManager|currentLocalNode" packages/movehat/src
(empty)

$ grep -RnE "\.mh\.runtime|getRuntime\(|getCurrentForkManager|getCurrentLocalNode|resetForkState|stopLocalTesting|teardownTestFixture" packages/movehat/src examples/counter-example
(empty)

$ grep -RnE "switchNetwork\(" packages/movehat/src examples/counter-example
(empty — confirms void→Promise<MovehatRuntime> change has no callers to migrate)

$ grep -nE "getMovehat|getRuntime|\\bmh\\b|mh\\.runtime" packages/movehat/src/__tests__/fixtures/sigint-deploy-harness.ts
(empty — harness uses Publisher / initRuntime directly)
```

## Note on `fixture.teardown()`

Deliberately does **not** call `AccountManager.clearPool()`. Two parallel `setupTestFixture()` calls share the static `AccountManager` pool today; if `teardown` cleared it, the first teardown would poison the second fixture's labeled accounts. The AccountManager-as-static-singleton concern lives in #55 separately and is not made worse by this PR — the pool just grows for the process lifetime (process exit reclaims it).

## Test plan

- [x] `pnpm --filter movehat exec tsc --noEmit` — clean
- [x] `pnpm --filter movehat test` — **184/184 passing**
- [x] `pnpm check:example` — clean (Tier 1 strict typecheck against `examples/counter-example` with `noUncheckedIndexedAccess` + `exactOptionalPropertyTypes` + `verbatimModuleSyntax`)
- [x] `pnpm test:example` — **9/9 passing** (Tier 2 runtime; pre-existing mocha teardown lag, same as M1.4 #100)
- [x] `pnpm test:smoke` — pass (Tier 2 install verification — packs + global install + CLI smoke)
- [N/A] `pnpm test:e2e` — Tier 3, only required pre-publish or for `develop → main` batch; not needed for sub-PRs landing in `develop`


gilbertsahumada (commented):

### 🔴 Blockers
None.

- **`src/templates/tests/Counter.test.ts:122`** — inline comment reads `// Cleanup: Stop local node and clear account pool` but `fixture.teardown()` deliberately does **not** clear the account pool (advisor catch #2). Users copy this file as scaffolding; a comment that misrepresents the behavior will mislead them when they hit the AccountManager-pool-still-populated edge case. Fix: trim the comment to `// Cleanup: Stop local node`.

### 🟢 Nits
- `examples/counter-example/tests/greeting.test.ts:80` has the same `// Cleanup: Stop local node` comment, which is now *accurate* — keep as-is.
- `LocalTestingContext` JSDoc mixes `@public` (top-level) with `@internal` (field-level) tags. TSDoc-style; api-extractor (M5 TypeDoc work) will respect both, but the combination is slightly unusual. Acceptable for now; revisit at M5.
- `@internal` JSDoc on `localNode? / forkServer? / forkManager?` is *documentation only* — these fields still appear in the public type. M5 TypeDoc + api-extractor will hide them from generated docs but won't suppress TS emit. The trade-off was deliberate: keep `forkManager.resetState()` reachable as an escape hatch without surface-area churn.
- Pre-existing mocha teardown lag during `pnpm test:example` resurfaces here, same as M1.4 #100. 9/9 still pass; `--exit: true` in `.mocharc.json` remains a tiny tracked follow-up (not part of this PR).

- `pnpm test:example` — **pass** (9/9, teardown lag pre-existing)
- `pnpm test:smoke` — **pass**
- `pnpm test:e2e` — N/A (sub-PR vs `develop`; Tier 3 runs pre-publish or on the `develop → main` batch PR per §6.3)

gilbertsahumada:
## Self-review resolution

**Resolved (in d365108):**
- 🟡 `src/templates/tests/Counter.test.ts:122` — trimmed misleading comment to `// Cleanup: Stop local node`.

**Deferred / not actioned:**
- 🟢 `LocalTestingContext` `@public` + field-level `@internal` mix — intentional, formalize at M5.
- 🟢 `@internal` JSDoc only documentation-level — intentional escape hatch for `forkManager.resetState()`.
- 🟢 Mocha teardown lag — pre-existing, follow-up `--exit: true` in `.mocharc.json` is tracked separately.

Ready to merge.

Issue #82, opened 2026-05-11
[M1.7] Strict types audit


gilbertsahumada:
Sub-issue of #68 (M1 — Testability refactors).

## Goal

`strict: true` is already enabled in `packages/movehat/tsconfig.json`, but the codebase predates that flag and likely still has `any`, non-null assertions, and unsafe casts.

## Definition of Done

- [ ] `grep -RnE ": any\\b| as any\\b|!\\." packages/movehat/src` audited; each remaining occurrence has a one-line comment explaining why or is replaced.
- [ ] Consider enabling `noUncheckedIndexedAccess` and `exactOptionalPropertyTypes`; if either causes too much churn, document the deferral in the issue closing comment.
- [ ] `pnpm --filter movehat build` stays green.

## Closes / Refs

Closes #57.
Tracks #68.

gilbertsahumada:
Closing during the §10 audit pass — M1.7 shipped as part of M1 (PR #106, develop → main batch). See [ROADMAP.md:66](https://github.com/gilbertsahumada/movehat/blob/main/ROADMAP.md#L66) for the M1 sub-PR table row: strict types audit (`any`, `!`, `noUncheckedIndexedAccess`). Closes audit issue #57 (`any`-in-fork bullet only; boundary validation deferred to follow-up #111).

GitHub's auto-close didn't fire originally because the sub-PR did not carry `Closes #82`. CLAUDE.md §10 (added in PR #249) sets the convention going forward.

Issue #46, opened 2026-04-24
[bug] switchNetwork does not return the new runtime


gilbertsahumada:
## Location
\`packages/movehat/src/runtime.ts:360-364\`

## Problem
\`switchNetwork\` calls \`initRuntime\` but discards the returned runtime. Callers that hold a reference to the old runtime keep stale \`aptos\` / \`account\` / \`config\` fields. Only the module-level \`cachedRuntime\` is updated (accessible via \`getRuntime()\`).

## Fix
- Make \`switchNetwork\` return \`Promise<MovehatRuntime>\`, OR
- Clearly document that users must re-fetch via \`getRuntime()\` / \`mh.runtime\` after switching.

## Acceptance
- [ ] Return signature reflects the new runtime, or docs updated
- [ ] Test showing correct network after switch

gilbertsahumada:
Satisfied retroactively by M1.5 (PR #107, commit `e464792`):

`runtime.switchNetwork` now has signature `(networkName: string) => Promise<MovehatRuntime>` (previously `Promise<void>`). The change was required by M1.5's removal of `cachedRuntime` — without the module-cached runtime to write back into, `switchNetwork` had to return the new runtime so callers could capture it.

Caller migration: the audit in PR #107 found zero callers of `switchNetwork` in `packages/movehat/src` or `examples/counter-example`, so the void → `Promise<MovehatRuntime>` change shipped with no consumer updates required. Anyone calling it externally must now do `const newRuntime = await runtime.switchNetwork(...)`.

Closed as part of M1.6 housekeeping (PR #109).

PR #109, opened 2026-05-13
feat(M1.6): cache loadUserConfig by mtime; fix #62 module leak (#81)

### LINKED ISSUE #62: [bug] Config loader cache-bust causes module memory leak
## Location
\`packages/movehat/src/core/config.ts:46-47, 53-54\`

## Problem
\`import(configUrl + '?t=' + Date.now())\` forces a fresh module per call. In a long-running test suite that calls \`loadUserConfig\` repeatedly, every call piles a new module into Node's loader cache → memory growth.

## Fix
Cache by \`{path, mtime}\` and reuse:
\`\`\`ts
const stat = fs.statSync(configPath);
const cacheKey = \`\${configPath}:\${stat.mtimeMs}\`;
\`\`\`
Only re-import when mtime changes. Document watch-mode behavior.

## Acceptance
- [ ] No new module created when config file unchanged
- [ ] Watch-mode still picks up edits (mtime changes)

gilbertsahumada:
## Summary

Replaces the per-call `?t=${Date.now()}` cache-bust in `loadUserConfig` with an in-memory map keyed by `{ absPath, mtimeMs }`. The parsed config is returned directly on subsequent calls when the file hasn't changed; the tsx loader register/unregister and the dynamic `import()` are skipped entirely on cache hits.

Closes **#62** (module-leak: every `loadUserConfig` call created a new module in Node's loader cache, which has no eviction).
Closes **#81** (M1.6).
Retroactively closes **#46** — the `switchNetwork`-returns-runtime piece already shipped in M1.5 (PR #107). Closing here as housekeeping; see comment posted on the issue after merge.

## DoD verification (issues #81 + #62)

| DoD bullet | Status |
|---|---|
| Cache by `{ absPath, mtimeMs }` | ✅ `Map<absPath, {mtimeMs, config}>` |
| No `?t=` query string remains | ✅ behavioral grep test in `config.test.ts:106` |
| Two consecutive `loadUserConfig(samePath)` hit disk **once** (logical-load) | ✅ ref-equality test in `config.test.ts:60` |
| Mtime modification invalidates the cache | ✅ `utimesSync`-based test in `config.test.ts:67` |
| No new module created when config file unchanged | ✅ cache hit skips `import()` entirely |
| Watch-mode picks up edits (mtime changes) | ✅ `?mtime=${mtimeMs}` makes Node see a new specifier when (and only when) mtime changes |

```bash
$ grep -nE "\\?t=['\"]\\s*\\+\\s*Date\\.now" packages/movehat/src/core/config.ts
(empty — and the behavioral test asserts this stays empty)

$ grep -nE "configCache|mtimeMs" packages/movehat/src/core/config.ts
(11 matches: declaration, lookup, comparison, miss-path import, write-back, reset)
```

## Why `?mtime=${mtimeMs}` instead of just dropping the query

Node's ESM loader caches modules by URL. With no query at all, `import(fileURL)` would return the same module forever per Node process — edits to the user's config would not be picked up at all (regression of the watch-mode use case in #62).

`?mtime=${mtimeMs}` makes Node treat each *distinct mtime* as a new specifier. The loader cache grows by the number of distinct edits (bounded), not by the number of `loadUserConfig` calls (unbounded under the old `?t=Date.now()`). And our own in-memory map short-circuits before `import()` is even invoked when mtime matches the cached entry — so on the common hot path, Node's loader is never re-consulted.

## Concurrency note

Two parallel `loadUserConfig()` calls on a cold cache may both invoke `import()`. Node's loader deduplicates by URL, so both resolve to the same module, and both writers race to store the same value into the Map. No corruption. An in-flight-promise memoization was considered and rejected as out-of-DoD-scope per CLAUDE.md §2 (no speculative complexity).

## Commits

- `0efa85c` feat(config): cache loadUserConfig by mtime, drop ?t= cache-bust
- `4176ef5` test(config): cover loadUserConfig mtime cache behavior + tick M1.6

## Test plan

- [x] `pnpm --filter movehat exec tsc --noEmit` — clean
- [x] `pnpm --filter movehat test` — **189/189 passing** (184 + 5 new)
- [x] `pnpm check:example` — clean (Tier 1 strict typecheck)
- [x] `pnpm test:example` — **9/9 passing** (Tier 2; pre-existing mocha teardown lag)
- [x] `pnpm test:smoke` — pass (Tier 2 install verification)
- [N/A] `pnpm test:e2e` — Tier 3, runs pre-publish or on the `develop → main` batch PR per CLAUDE.md §6.3

## Remaining for M1 batch to `main`

Only **M1.7** (#82, strict types audit) left. Then `develop → main` batch with `test:e2e` gate + README sync.

### DIFF
--- a/ROADMAP.md
+++ b/ROADMAP.md
@@ -63,7 +63,7 @@ Each milestone below lists **explicit, mechanically verifiable** acceptance crit
 | ✅ | M1.3c | [#78](https://github.com/gilbertsahumada/movehat/issues/78) (shipped in PR #97) | Migrate `runtime.ts` publish + `node/LocalNodeManager.ts` daemon + stderr-redaction unit test | completes [#58](https://github.com/gilbertsahumada/movehat/issues/58), completes [#43](https://github.com/gilbertsahumada/movehat/issues/43) |
 | ✅ | M1.4 | [#79](https://github.com/gilbertsahumada/movehat/issues/79) | Extract `core/Publisher.ts` + per-deploy unique profile + signal handler | [#19](https://github.com/gilbertsahumada/movehat/issues/19), [#36](https://github.com/gilbertsahumada/movehat/issues/36), [#37](https://github.com/gilbertsahumada/movehat/issues/37), [#38](https://github.com/gilbertsahumada/movehat/issues/38), [#53](https://github.com/gilbertsahumada/movehat/issues/53) |
 | ✅ | M1.5 | [#80](https://github.com/gilbertsahumada/movehat/issues/80) | Remove `cachedRuntime` and the three `setupLocalTesting` singletons; `switchNetwork` returns runtime (folded in from M1.6) | [#21](https://github.com/gilbertsahumada/movehat/issues/21), [#55](https://github.com/gilbertsahumada/movehat/issues/55) |
-| ⏳ | M1.6 | [#81](https://github.com/gilbertsahumada/movehat/issues/81) | `loadUserConfig` mtime cache (the `switchNetwork`-returns-runtime piece shipped with M1.5) | [#46](https://github.com/gilbertsahumada/movehat/issues/46), [#62](https://github.com/gilbertsahumada/movehat/issues/62) |
+| ✅ | M1.6 | [#81](https://github.com/gilbertsahumada/movehat/issues/81) | `loadUserConfig` mtime cache (the `switchNetwork`-returns-runtime piece shipped with M1.5) | [#46](https://github.com/gilbertsahumada/movehat/issues/46), [#62](https://github.com/gilbertsahumada/movehat/issues/62) |
 | ⏳ | M1.7 | [#82](https://github.com/gilbertsahumada/movehat/issues/82) | Strict types audit (`any`, `!`, `noUncheckedIndexedAccess`) | [#57](https://github.com/gilbertsahumada/movehat/issues/57) |
 
 **Definition of Done** (rolled up from the sub-issues):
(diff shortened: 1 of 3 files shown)


gilbertsahumada (commented):

### 🔴 Blockers
None.

None.

### 🟢 Nits
- **Test count: 189 vs the plan's 190.** Test 6 from the plan (`_resetConfigCache forces a re-read`) was dropped during implementation because the impl revealed it wasn't behaviorally meaningful: after `_resetConfigCache()`, the next call still hits Node's URL-keyed loader cache for the same `?mtime=` URL and returns the same module reference. The cache function is still test-useful (it gives `beforeEach` isolation; if it didn't work, tests 3+4 would leak state) — its coverage is implicit in the rest of the suite continuing to pass. Test count went 184 → 189, ROADMAP + PR body reflect this consistently.
- **Source-grep test pattern (`config.test.ts:106`)** is `/\?t=['"]\s*\+\s*Date\.now/` — strict enough that the explanatory comment at `config.ts:13` ("the previous `?t=Date.now()` cache-bust") doesn't false-positive. If someone deletes that comment, the test still passes. Pattern matches the *concatenation*, not the literal substring `?t=`.
- **Underscore-prefix on `_resetConfigCache`** is the conventional "this exists but you shouldn't depend on it" marker. The `@internal` JSDoc plus the relative-path-only import from the test file gives layered defense; eventual api-extractor work at M5 will formalize.
- **Concurrency: two cold-cache `loadUserConfig()` calls race → both call `import()`.** Documented in the cache declaration comment. Node's URL-keyed loader cache deduplicates, both writers store the same value into the Map, no corruption. Not memoizing the in-flight Promise was a deliberate CLAUDE.md §2 simplicity call — the cost of the duplicate `import()` on cold-cache contention is bounded (one extra parse on a rare race).
- **Mocha teardown lag** during `pnpm test:example` resurfaces here, same as M1.4/M1.5. 9/9 still pass; `--exit: true` in `.mocharc.json` is still a tiny follow-up not part of this PR.

- `pnpm test:example` — **pass** (9/9). First run flaked due to leftover Movement process from prior M1.5 smoke test; `pkill movement / lsof -ti:8080,8081,8070 | xargs kill` cleared the state. Not a regression from this PR — same Movement-state hygiene issue we hit in M1.4/M1.5.
- `pnpm test:smoke` — **pass**
- `pnpm test:e2e` — N/A (sub-PR vs `develop`; Tier 3 runs on `develop → main` batch per §6.3)

PR #110, opened 2026-05-13
feat(M1.7): strict types audit + flag enablement (#82, #57)

gilbertsahumada:
## Summary

Closes the M1 testability-refactor milestone with a strict TypeScript types audit:

1. **Enable two strict flags** in `packages/movehat/tsconfig.json` — `noUncheckedIndexedAccess` + `exactOptionalPropertyTypes` — closing the gap with `examples/counter-example/tsconfig.json` (which has carried these flags as the strictness gate per CLAUDE.md §6.1).
2. **Audit the 59-instance `any` surface and the 36-instance `!` surface** identified in Phase 1. Replace where cheap, justify with a one-line adjacent comment where propagation cost would exceed the safety gain.

Closes **#82** (M1.7). Closes the `any`-in-fork bullet of **#57** — the boundary-validation bullet (zod or hand-written guards for `fork/api.ts` parsed JSON) is **deferred** to a follow-up sub-issue, filed at merge time per the locked plan decision.

## Flag enablement (Commit 1: `edc6fa4`)

`packages/movehat/tsconfig.json` previously had only `"strict": true`. Now also:

```diff
+    "noUncheckedIndexedAccess": true,
+    "exactOptionalPropertyTypes": true,
```

**`verbatimModuleSyntax` deliberately omitted.** Its safety benefit is small for movehat (no isolated-modules build constraints) and the mechanical migration would touch 30+ type-only imports for negligible gain.

**Surface fixed:**

- `noUncheckedIndexedAccess` (~10 sites): regex `match[N]` captures, accounts/array indexing in `runtime.ts` / `config.ts`, dev-address table walks in `commands/compile.ts`, pathname splits in `fork/server.ts`, gradient palette walk in `ui/colors.ts`. Pattern: capture into a local + guard, or `?? ""` where the impossible-undefined branch is a no-op (e.g. `String.split` always returns ≥1 element).
- `exactOptionalPropertyTypes` (~10 sites): widened `RunCliOptions.adapter`, `PublisherDeps.adapter`, `PublishInput.packageDir`, `DeploymentInfo.txHash/blockNumber`, `RunMoveTestsOptions.{filter,ignoreWarnings,skipIfMissing}`, `StoredAccount.label` to `T | undefined`. Restructured `npm-registry.ts` fetch call to conditionally pass `signal`.

### Replaced
[...]

PR #106, opened 2026-05-13
[M1 batch] Testability refactors → main (#68)

gilbertsahumada:
## Summary

**M1 milestone batch merge** — ships the testability-refactor milestone (#68) plus adjacent docs-branding and housekeeping work that landed on `develop` since the last batch (#102, merged 2026-05-13 morning).

This is a **main-affecting merge** — requires explicit user approval before I'll click merge.

### M1 sub-PRs (#76, #87, #89, #92, #97, #100, #107, #109, #110)

| Sub-PR | Issue | Title | Lands |
|---|---|---|---|
| M1.1 | foundations | `utils/{runCli,childProcessAdapter,address}.ts` + tests | PR #76 |
| M1.2 | #56 | Migrate `fork/{manager,api,storage}.ts` to `utils/address.ts` | PR #87 |
| M1.3a | foundations | Extend `ChildProcessAdapter` with `spawn()` + `inheritStdio` | PR #89 |
| M1.3b | partial #58 | Migrate 6 simple callsites to `runCli` | PR #92 |
| M1.3c | #58, #43 | Migrate `runtime.ts` publish + `LocalNodeManager.ts` daemon | PR #97 |
| M1.4 | #19, #36, #37, #38, #53 | Extract `core/Publisher.ts` + unique profile + signal handler | PR #100 + #101 + #102 + #103 (CodeRabbit fixes) |
| M1.5 | #21, #55 | Kill `cachedRuntime` + 3 setupLocalTesting singletons; `switchNetwork` returns runtime | PR #107 |
| M1.6 | #62 | `loadUserConfig` mtime cache (#46 retroactively closed) | PR #109 |
| M1.7 | #82, partial #57 | Strict-types audit + `noUncheckedIndexedAccess` + `exactOptionalPropertyTypes` | PR #110 |

### Closed by this batch
- #43 (stderr secret redaction), #19 (Publisher extraction), #36 (SIGINT cleanup), #37 (parallel-deploy yaml mutex), #38 (Move.toml immutability), #53 (per-deploy unique profile)
- #56 (address normalization), #58 (runCli migration)
- #21 (cachedRuntime), #55 (setupLocalTesting singletons), #46 (switchNetwork returns runtime — retroactive)
- #62 (config module-leak), #82 (strict types), #57 partial (any-in-fork; boundary-validation deferred to #111)

### Adjacent non-M1 work also shipping

[...]

PR #114, opened 2026-05-13
fix: address high-priority CodeRabbit findings on PR #106 (TP1-TP4 + TP7)

gilbertsahumada:
## Summary

Addresses the **high-priority** CodeRabbit findings from PR #106 (the M1 \`develop → main\` batch). Pulls in 5 actionable bug fixes + a ROADMAP cleanup. The 3 false-positive CodeRabbit comments get inline responses on PR #106 (not on this PR — the comments live on #106).

## Commits

| Commit | TP | Issue |
|---|---|---|
| \`0d12173\` | **TP1** 🟠 | \`setupWithLocalNode\` / \`setupWithFork\` leak the started child if post-start steps throw. Wrap in try/catch; \`stop()\` on error. |
| \`a868f00\` | **TP2** 🟠 | Same leak at fixture-assembly level (\`setupTestFixture\` / \`setupMinimalFixture\`). Call \`ctx.teardown()\` on error. |
| \`0cdcef1\` | **TP3** 🟠 | Post-publish errors (\`saveDeployment\`, inner-finally \`removeProfile\`) misreported as publish failures. New \`PostPublishError\` class carries the successful \`DeploymentInfo\`; inner-finally now \`.catch(...)\` instead of throwing. |
| \`993f7dc\` | **TP7** 🟡 | \`fork/manager.ts\` + \`fork/server.ts\` (5 sites): \`: ""\` fallback for non-Error throws → \`: String(error)\`. |
| \`5d81248\` | **TP4** 🟡 | \`ROADMAP.md\` placeholder \`#N\` → \`#110\` + add test:e2e evidence reference. |

## Why TP1 + TP2 matter

We've hit the orphan-process bug repeatedly during Tier-2 runs across M1.5, M1.6, and M1.7 — manifests as "Movement command failed" on a second \`test:example\` invocation, because the first run's leaked node still owns port 8080. Today the workaround is \`pkill -f movement && lsof -ti:8080 | xargs -r kill -9\` between runs. After TP1+TP2, a fixture-assembly failure stops the node automatically; no leak, no port conflict on the next run.

## Why TP3 matters

The publish flow had two latent bugs:

1. **\`await\` in finally clobbering the publish result.** If \`removeProfile\` failed in the inner \`finally\`, the throw replaced both successful returns and outer errors. Successful deploys could look failed; real publish errors could be masked.
2. **Post-publish bookkeeping misreported as publish failure.** Once the on-chain transaction succeeded, any later throw (saveDeployment write, etc.) bubbled out as "Failed to publish module" — inviting the user to redeploy and burn gas on a duplicate transaction.

The fix introduces a clear "publish succeeded" boundary (the existing \`logger.success("Module published successfully!")\` line). Anything thrown below that line is now a \`PostPublishError\` carrying the \`DeploymentInfo\` so callers can recover (write the JSON manually, or trust the on-chain state).

## False positives — inline responses on PR #106

Three of CodeRabbit's findings were verified as false positives and will get inline replies on PR #106 explaining why:

- **FP1** \`setupLocalTesting.ts:307-325\` (resetForkState uses console.warn) — function was deleted in M1.5 (PR #107). CodeRabbit reviewed an older base.
- **FP2** \`deployContract.test.ts:15\` 🔴 (raw spawn import) — test code spawning its own SIGINT harness; adapter pattern is for production code.
- **FP3** \`ui/symbols.ts:20-25\` (\`step\` hardcoded \`'▸'\`) — intentional design choice documented in adjacent comment.

## Deferred to follow-up sub-PR C (out of scope here)

7 low-priority CodeRabbit findings deferred to a post-batch polish sub-PR:
- TP5: Signal handler unregister (Publisher.ts:183) — embedded-library concern; movehat is CLI today.
- TP6: \`fixture.teardown()\` undefined guard in tests — already covered if TP1/TP2 ship (root cause is the setup throw).
- TP8: Empty palette guard in \`ui/colors.ts\`.
- TP9: \`console.log\` → \`logger.plain\` in \`LocalNodeManager.ts:67\` + \`view-resource.ts:42-43\`.
- TP10: \`structuredClone\` defensive copy in config cache (debatable — defeats cache speedup).
- TP11: \`fetch\` / \`fs.readFile\` error handling in docs site routes.

## Test plan

- [x] \`pnpm --filter movehat exec tsc --noEmit\` — clean
- [x] \`pnpm --filter movehat test\` — **189/189 passing** after every commit
- [x] \`pnpm check:example\` — clean
- [x] \`pnpm test:example\` — **9/9** (Tier 2; pre-existing mocha teardown lag)
- [x] \`pnpm test:smoke\` — pass
- [N/A] \`pnpm test:e2e\` — Tier 3 runs on PR #106 (the batch PR); this sub-PR auto-updates it.

## Public-API note

New exported class \`PostPublishError\` joins the existing \`ModuleAlreadyDeployedError\` in \`movehat\`'s public barrel. Pre-1.0; goes into M6 release notes when 0.1.0 ships.

### DIFF
--- a/ROADMAP.md
+++ b/ROADMAP.md
@@ -79,7 +79,7 @@ Each milestone below lists **explicit, mechanically verifiable** acceptance crit
 - [x] `examples/counter-example/` keeps passing through every sub-PR (per Decision 6) — mechanical typecheck gate enforced by pre-push since PR #84; runtime gate **9/9 passing** as of PR closing #86 (`message.test.ts` rewritten to local-node, broken-scaffold `greeting-fork.test.ts` removed)
 - [x] M1.7 (#82): `noUncheckedIndexedAccess` + `exactOptionalPropertyTypes` enabled in `packages/movehat/tsconfig.json`; all 16 `catch (error: any)` sites migrated to `unknown`-typed catches; `fork/*` `any` audit closes the corresponding bullet of #57 — boundary validation (zod or guards for `fork/api.ts` parsed JSON) deferred to follow-up; `verbatimModuleSyntax` deliberately omitted (low-benefit/high-churn)
 
-**M1 status: ✅ shipped via PRs #76, #87, #89, #92, #97, #99 (M1.4), #107 (M1.5), #109 (M1.6), and #N (M1.7). Ready for `develop → main` batch.**
+**M1 status: ✅ shipped via PRs #76, #87, #89, #92, #97, #99 (M1.4), #107 (M1.5), #109 (M1.6), and #110 (M1.7). Tier 3 verification `pnpm test:e2e` 32/32 passing (captured on the `develop → main` batch PR #106). Ready for `develop → main` batch.**
 
 ### M2 — Hardhat-style Harness API (~6 days, issue #69)
 
--- a/packages/movehat/src/core/Publisher.ts
+++ b/packages/movehat/src/core/Publisher.ts
@@ -22,7 +22,7 @@ import {
   validateSafeName,
 } from "./deployments.js";
 import { validatePathSafety, validateProfileSafety } from "./shell.js";
-import { CliExecutionError, ModuleAlreadyDeployedError } from "../errors.js";
+import { CliExecutionError, ModuleAlreadyDeployedError, PostPublishError } from "../errors.js";
 import { runCli } from "../utils/runCli.js";
 import { logger } from "../ui/index.js";
 import type { ChildProcessAdapter } from "../utils/childProcessAdapter.js";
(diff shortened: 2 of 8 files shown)


gilbertsahumada (commented):

### 🔴 Blockers
None.

- **No dedicated unit test for the TP3 PostPublishError path.** \`saveDeployment\` failure is hard to trigger without filesystem mocking that the existing \`__tests__/deployContract.test.ts\` harness doesn't have (it uses real \`tmpdir\` for \`~/.aptos\`). The mechanical fix is clear; tracked as a follow-up for sub-PR C. Without this test, a future regression that re-collapses post-publish errors into "Failed to publish" would only surface in production.
- **No dedicated unit test for TP1/TP2 cleanup-on-failure.** The integration test is "run \`pnpm test:example\` twice back-to-back without manual port cleanup" — today the second run flakes, after this PR it doesn't. The regression would manifest at Tier 2, not Tier 1. Acceptable for now; \`helpers/__tests__/\` is greenfield infra.

### 🟢 Nits
- \`PostPublishError\` re-export from \`src/index.ts\` adds public API surface. Pre-1.0 — documented in PR body. Will appear in M6 release notes.
- The inner-finally \`.catch\` swallows the cleanup error to a \`logger.warning\`. We deliberately do NOT throw because throwing in a finally clobbers the publish result/error. If someone wants the cleanup failure to also propagate, they'd need to refactor toward an explicit "aggregate error" pattern — out of scope.
- \`PostPublishError.cause\` shadows the native \`Error.cause\` (ES2022). Chose explicit field for clarity in the typed public surface; the redundancy is harmless and improves discoverability.
- The recovery instruction in the warning log names the file path (\`deployments/{network}/{moduleName}.json\`) — useful for users hitting this edge case, but implicitly couples the warning to the current \`saveDeployment\` storage layout. If we change the storage shape, update the warning string.

- \`pnpm --filter movehat test\` — 189/189 after every commit
- \`pnpm check:example\` — clean
- \`pnpm test:example\` — **9/9** (Tier 2; pre-existing mocha teardown lag same as M1.5-M1.7)
- \`pnpm test:smoke\` — pass

Ready to merge.

PR #115, opened 2026-05-13
chore: reference GRANT.pdf from ROADMAP + gitignore *.pdf

gilbertsahumada:
## Summary

Two coupled housekeeping changes:

1. **\`.gitignore\`**: append \`*.pdf\` to keep local-only reference documents (the MOU \`GRANT.pdf\`, plus any future PDF material) out of the public repo. The file is currently in the working tree but never staged; this rule prevents accidental commits.
2. **\`ROADMAP.md\`**: add a "Reference documents" subsection near the top pointing at \`GRANT.pdf\` (local-only) + annotate every milestone header with the corresponding KPI tag (\`(KPI 1)\` / \`(KPI 2)\` / \`(prerequisite for KPI 1)\`) so reviewers can trace each milestone to the underlying grant deliverable.

## Why

The roadmap maps directly to deliverables defined in the MOU (\`GRANT.pdf\`). Until now the mapping was implicit. Adding the file reference + per-milestone KPI tags makes the link explicit without mirroring the MOU's specific terms (dollar amounts, vesting schedules, parties) into the public repo.

## Privacy guardrail

No financial terms, vesting schedules, party names, or contractual text from the MOU appear in this PR. Only the filename (\`GRANT.pdf\`) and the KPI labels (\`KPI 1\`, \`KPI 2\`) are referenced.

## KPI mapping (per the labels added)

| Milestone | KPI |
|---|---|
| M0 ✅ | KPI 2 |
| M1 ✅ | prerequisite for KPI 1 |
| M2 | KPI 1 |
| M3 | KPI 1 |
| M4 | KPI 1 |
| M5 | KPI 2 |
| M6 | KPI 2 |
| M7 | KPI 2 |

## Test plan

- [x] \`git check-ignore -v GRANT.pdf\` → matches \`.gitignore:56 *.pdf\` rule
- [x] No code changes; Tier 1/2 not required.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -53,3 +53,4 @@ pnpm-debug.log*
 !MOVEMENT_CLI_COMPAT.md
 !NOTICE.md
 !.github/**/*.md
+*.pdf
--- a/ROADMAP.md
+++ b/ROADMAP.md
@@ -2,6 +2,10 @@
 
 This roadmap organizes the next phase of Movehat development. Each milestone has explicit Definition of Done criteria and is tracked as a meta-issue on GitHub.
 
+## Reference documents
+
+- **`GRANT.pdf`** (local-only, gitignored) — underlying MOU that defines the deliverable KPIs this roadmap maps to. Kept out of the repo by the `*.pdf` rule in `.gitignore`. Milestone headers below carry `(KPI 1)` / `(KPI 2)` tags pointing at the relevant KPI section in `GRANT.pdf`. When in doubt about acceptance criteria, the MOU text is the source of truth.
+
 ## Current state
 
 - Documentation site shipped (Fumadocs + Next.js, static export)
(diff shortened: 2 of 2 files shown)


gilbertsahumada (commented):
Self-review per CLAUDE.md §8:

🔴 Blockers: none.
🟡 Worth fixing: none.
🟢 Nits:
- The `*.pdf` rule is broad — it would also ignore future tracked PDFs (benchmarks, architecture diagrams, etc.) if any ever land in the repo. If that becomes a need later, add a specific `!path/to/wanted.pdf` exception. Today there are zero tracked PDFs, so the broad rule is fine.
- The KPI labels are intentionally vague (`(KPI 1)` / `(KPI 2)` instead of `(KPI 1: Forklift API + Testing)`). A reviewer with the MOU in hand can decode; the public repo doesn't leak the breakdown. Deliberate trade-off between traceability and privacy.

Ready to merge.

Issue #119, opened 2026-05-13
[M2.4] Migrate example + templates + docs sync; deprecate mh()


gilbertsahumada:
Sub-issue of #69 (M2 — Hardhat-style Harness API). Depends on M2.3 landing first.

## Goal
Switch the install-experience to Harness: one example test, the published template, and 8 docs pages. Mark `mh()` / `getMovehat` / `initRuntime` with `@deprecated` JSDoc. Behavior unchanged.

### Migration
- [ ] `examples/counter-example/tests/Counter.test.ts` uses `Harness.createLocal()`
- [ ] `examples/counter-example/scripts/deploy-counter.ts` uses `Harness.createLive` + `deployCodeObject`
- [ ] `greeting.test.ts` + `message.test.ts` stay on `setupTestFixture` (coexistence demo)
- [ ] `packages/movehat/src/templates/scripts/deploy-counter.ts` rewritten to Harness
- [ ] `packages/movehat/src/templates/movehat.config.ts` JSDoc mentions Harness as recommended entry

### Deprecation (non-breaking)
- [ ] `@deprecated` JSDoc on `getMovehat`, `initRuntime` in `packages/movehat/src/runtime.ts`
- [ ] `@deprecated` JSDoc on `mh`/`getMovehat`/`initRuntime` re-exports in `packages/movehat/src/index.ts`

### Docs
- [ ] 8 MDX files migrated (`index`, `cli/init`, `cli/run`, `getting-started/{quickstart,configuration}`, `guides/{testing,scripts,deployment}`)
- [ ] New: `packages/docs/content/docs/api/harness.mdx` covers 7 methods + cleanup
- [ ] `packages/docs/content/docs/api/meta.json` updated/created so harness.mdx shows in nav
- [ ] `pnpm build:docs` clean
- [ ] `grep -rE '\bmh\(|getMovehat|setupTestFixture' packages/docs/content/docs/{index,getting-started,guides,api}` returns only the deprecation note

### Verification (full ladder)
- [ ] `pnpm --filter movehat exec tsc --noEmit` clean
- [ ] `pnpm check:example` Tier 1 green
- [ ] `pnpm test:example` Tier 2 green
- [ ] `pnpm test:smoke` Tier 2 green
- [ ] `pnpm test:e2e` Tier 3 green (templates changed → new-user flow must work)

Tracks #69. After this lands, all M2 DoD bullets in ROADMAP.md flip to `[x]` and the `develop → main` batch PR can be opened.

Issue #71, opened 2026-05-08
[Roadmap] M4: Zero-mock integration suite + E2E SLO


gilbertsahumada:
## Goal
Build an integration test suite running the **real** Movement CLI; tighten CI E2E policy to run on every push within a strict 5-minute budget.

### Integration suite
- [ ] `packages/movehat/test/integration/` directory exists
- [ ] Suite drives the full Harness flow at least once: `createLocal` → `deployCodeObject` → `runViewFunction` → `upgradeCodeObject` → `runMoveScript` → `cleanup`
- [ ] At least one test path uses `Harness.createFork(...)`
- [ ] `grep -r "vi\.mock" packages/movehat/test/integration/` returns **no matches**
- [ ] Suite runs via a separate `vitest.integration.config.ts`
- [ ] Helper that asserts Movement CLI is installed; skips suite locally with a clear message if not

### CI policy
- [ ] `.github/workflows/ci.yml` E2E job triggers on **every push** (`on: push`), not only on PR to `main`
- [ ] `timeout-minutes: 5` set per job
- [ ] Wall-clock for the E2E job is **<5 minutes** on cache hit (verified with at least 3 consecutive runs)
- [ ] CI green on `main` and on at least one non-`main` branch (e.g., a feature/* or security/* branch active during M4)

### Documentation
- [ ] `TESTING.md` section: "Running the integration suite locally"
- [ ] `TESTING.md` section: "Running the suite in Docker" using existing `docker-compose.test.yml`
- [ ] `TESTING.md` section: "How to skip integration tests during dev"

## Verification commands
```bash
# Zero mocks in integration:
grep -rn "vi\.mock" packages/movehat/test/integration/

# CI policy:
yq '.on' .github/workflows/ci.yml      # should include 'push'
yq '.jobs.*."timeout-minutes"' .github/workflows/ci.yml | sort -u    # should be <=5

# CI green:
gh run list --branch main --limit 5 --json conclusion --jq '.[].conclusion' | grep -v success || echo "all green"
```

## Effort
~4 days

## Dependencies
Requires #69 (M2 Harness API) landed; #70 (M3 cache) lands in parallel.

## Out of scope
- Coverage of the integration suite (it is not measured against the 80% threshold).

## Artifacts
2 PRs:
1. `test: zero-mock integration suite`
2. `ci: E2E on every push, 5-minute timeout SLO`

Issue #69, opened 2026-05-08
[Roadmap] M2: Hardhat-style Harness API with cleanup poisoning


gilbertsahumada:
## Goal
Provide a Hardhat-style testing harness with explicit lifecycle and use-after-cleanup safety. New public-facing API for tests and programmatic use.

### API surface (each method must be callable from the built package)
- [ ] `import { Harness } from 'movehat'` works from `dist/`
- [ ] `Harness.createLocal()` returns a usable instance
- [ ] `Harness.createFork(network: string, apiKey?: string)` returns a usable instance
- [ ] `Harness.createLive(network: string, faucetUrl?: string)` returns a usable instance
- [ ] `harness.deployCodeObject(options)` deploys a real Move package via `move deploy-object`
- [ ] `harness.upgradeCodeObject(options)` upgrades an existing code object
- [ ] `harness.runViewFunction(options)` executes a read-only view function
- [ ] `harness.runMoveScript(options)` compiles and executes a Move script
- [ ] `harness.cleanup()` releases the local node + temp dirs + sets `poisoned = true`

### Use-after-cleanup safety
- [ ] Calling **any** method on a disposed harness (other than `cleanup`) throws `HarnessDisposedError` synchronously
- [ ] `HarnessDisposedError` exported and documented

### Migration
- [ ] `examples/counter-example/` uses `Harness.createLocal()`
- [ ] `packages/movehat/src/templates/scripts/deploy-counter.ts` uses the new API
- [ ] `packages/movehat/src/templates/movehat.config.ts` updated as needed
- [ ] All MDX code snippets in `packages/docs/content/docs/{cli,guides,getting-started}/` updated
- [ ] `packages/docs/content/docs/api/harness.mdx` exists and covers the 7 methods + cleanup
- [ ] `mh()` still exported with `@deprecated` JSDoc

### Tests
- [ ] New unit tests cover: factory methods (3), each method (4), proxy poisoning, cleanup idempotency
- [ ] `pnpm test` green
- [ ] `pnpm build` clean (no TS errors on the new module)

## Implementation layout

`packages/movehat/src/harness/`:
- `Harness.ts` — class with static factory methods
- `codeObject.ts` — `deployCodeObject`, `upgradeCodeObject`
- `script.ts` — `runMoveScript`
- `view.ts` — `runViewFunction`
- `proxy.ts` — generic Proxy helper poisoning instances after `cleanup()`
- `errors.ts` — `HarnessDisposedError`
[...]

PR #122, opened 2026-05-13
feat(harness): runViewFunction + runMoveScript + parseTxHash extract (M2.3)

### LINKED ISSUE #118: [M2.3] runViewFunction + runMoveScript
Sub-issue of #69 (M2 — Hardhat-style Harness API). Depends on M2.2 landing first.

## Goal
Implement the read-side and script-side Harness methods. `runViewFunction` reuses the SDK call already in `MoveContract.view`. `runMoveScript` wraps `movement move run-script --compiled-script-path` via `runCli`.

### API
- [ ] `harness.runViewFunction({ function, typeArgs?, args? })` returns `Promise<unknown[]>` (view-function return tuple); delegates to `this.runtime.aptos.view()`
- [ ] `harness.runMoveScript({ scriptPath, args?, typeArgs?, privateKey? })` returns `Promise<MoveScriptResult>` with parsed `txHash`
- [ ] `runMoveScript` throws clear error on fork harness; `runViewFunction` works on fork

### Types
- [ ] `packages/movehat/src/types/harness.ts` extended with `RunViewFunctionOptions`, `RunMoveScriptOptions`, `MoveScriptResult`

### Tests
- [ ] `harness/view.test.ts` — mocked `aptos.view`; correct payload shape
- [ ] `harness/script.test.ts` — fake `ChildProcessAdapter`; txHash parse + error path
- [ ] `pnpm --filter movehat test` green
- [ ] Tier 1 + Tier 2

### Files
- Create: `packages/movehat/src/harness/{view,script}.ts`
- Edit: `packages/movehat/src/harness/Harness.ts` (replace remaining M2.1 stubs)
- Edit: `packages/movehat/src/types/harness.ts`

Tracks #69.

gilbertsahumada:
## Summary

Third sub-PR of M2 (#69). Closes the seven-method Harness API surface required by KPI 1: `runViewFunction` (SDK delegation) and `runMoveScript` (CLI wrapper) replace the last two M2.1 stubs.

## Commits

1. **`refactor(utils): extract parseTxHash to utils/parseCliOutput.ts`** — the same regex pair (transaction-hash context + 64-char hex fallback) lived in 2 places after M2.2 and M2.3 needed a third. Lift once; Publisher.ts + harness/codeObject.ts + harness/script.ts import. Behavior-neutral: **210/210 tests pass unchanged**.

2. **`feat(harness): runViewFunction + runMoveScript + tests`** — adds:
   - `harness/view.ts` — 6-line wrapper around `runtime.aptos.view(...)`. No CLI invocation, no profile management. Returns raw `unknown[]` (SDK contract) — callers destructure: `const [count] = await harness.runViewFunction({...})`. Works on all 3 modes including fork (read-only).
   - `harness/script.ts` — wraps `movement move run-script` via `runCli` + shared `movementProfile` helpers (same security model as Publisher / codeObject: per-deploy unique profile, atomic 0o600 yaml writes, SIGINT-safe sync cleanup, `--profile` auth). **Auto-detects script kind from extension**: `.move` → `--script-path` (CLI compiles inline), `.mv` → `--compiled-script-path`. Throws synchronously on unsupported extensions or missing files. `txHash` parsed via shared helper; `success` + `vmStatus` best-effort from the Result JSON block.
   - `types/harness.ts` — new `RunViewFunctionOptions`, `RunMoveScriptOptions`, `MoveScriptResult`.
   - `Harness.ts` — replaces both stubs. `runMoveScript` has fork-mode guard (write operation); `runViewFunction` does not.
   - 11 new test cases (4 view + 7 script).
   - Retires the M2.1 \"stubs reject before cleanup\" test (all 4 methods now real).

## Key design decisions

- **`runMoveScript` accepts both `.move` and `.mv`**: the CLI supports both natively; exposing both costs ~5 LoC and matches the `movehat compile` workflow. Documented in JSDoc.
- **`runViewFunction` returns raw `unknown[]`**: matches the SDK's `aptos.view` contract. No singleton unwrap (the existing `MoveContract.view` does unwrap, but Harness is the new primary API and predictable typing wins). Users destructure for singletons.
- **`runViewFunction` mocking strategy**: tests monkey-patch `harness.runtime.aptos.view = vi.fn()...` because the SDK isn't behind an injectable adapter. One-off for M2.3; the wrapper is a 6-line forward. Adapter-injecting the entire Aptos SDK is way out of scope.

## DoD ticked in ROADMAP

- [x] `harness.runViewFunction(options)` executes a view function
- [x] `harness.runMoveScript(options)` compiles and executes a Move script

All 7 API methods are now implemented. Remaining M2 work is migration + docs in M2.4 (#119).

## Test plan

- [x] Tier 1: `pnpm --filter movehat exec tsc --noEmit` — clean
- [x] Tier 1: `pnpm --filter movehat test` — **220/220 passing** (210 prior + 11 new − 1 retired stub-rejection assertion)
- [x] Tier 1: `pnpm check:example` — green
- [x] Tier 2: `pnpm test:smoke` — pass (mandatory: types surface changed with 3 new exports; tarball + global install + CLI sanity confirmed)
- [skip] Tier 2: `pnpm test:example` — same mocha teardown flake documented in M2.2 PR description. Publish path is untouched at the regex level (C1's parseTxHash extraction is byte-identical to the prior inline pattern), so the integration angle is a strict subset of what M4 (#71) will cover.
- [N/A] Tier 3: `pnpm test:e2e` — templates + CLI untouched in M2.3; Tier 3 runs at M2.4.

Tracks #69. Closes #118.

### DIFF
--- a/ROADMAP.md
+++ b/ROADMAP.md
@@ -105,8 +105,8 @@ Each milestone below lists **explicit, mechanically verifiable** acceptance crit
 - [x] `Harness.createLive(network: string, faucetUrl?: string)` returns a usable instance (M2.1)
 - [x] `harness.deployCodeObject(options)` deploys a real Move package (M2.2 — wraps `movement move deploy-object`; reuses Publisher hardening via the extracted `core/movementProfile.ts` helpers; integration coverage in M4)
 - [x] `harness.upgradeCodeObject(options)` upgrades an existing code object (M2.2 — wraps `movement move upgrade-object`; same shared helpers)
-- [ ] `harness.runViewFunction(options)` executes a view function
-- [ ] `harness.runMoveScript(options)` compiles and executes a Move script
+- [x] `harness.runViewFunction(options)` executes a view function (M2.3 — delegates to `aptos.view`; returns raw `unknown[]`; works on all 3 harness modes including fork since views are read-only)
+- [x] `harness.runMoveScript(options)` compiles and executes a Move script (M2.3 — wraps `movement move run-script`; auto-detects `.move` vs `.mv` from extension; `parseTxHash` shared via `utils/parseCliOutput.ts`)
 - [x] `harness.cleanup()` releases the local node, removes temp dirs, and sets `poisoned = true` (M2.1 — idempotent; stops owned localNode / forkServer; tested)
 
 **Definition of Done — Use-after-cleanup safety**:
--- a/packages/movehat/src/__tests__/harness/Harness.proxy.test.ts
+++ b/packages/movehat/src/__tests__/harness/Harness.proxy.test.ts
@@ -99,8 +99,12 @@ version = "0.0.1"
     expect(() =>
       harness.upgradeCodeObject({ moduleName: "x", objectAddress: "0x1" })
     ).toThrow(HarnessDisposedError);
-    expect(() => harness.runViewFunction({})).toThrow(HarnessDisposedError);
-    expect(() => harness.runMoveScript({})).toThrow(HarnessDisposedError);
+    expect(() =>
+      harness.runViewFunction({ function: "0x1::m::f" })
+    ).toThrow(HarnessDisposedError);
+    expect(() =>
+      harness.runMoveScript({ scriptPath: "irrelevant.move" })
+    ).toThrow(HarnessDisposedError);
   });
 
   it("HarnessDisposedError carries the offending method name", async () => {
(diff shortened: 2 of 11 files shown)


gilbertsahumada (commented):
## Self-review (CLAUDE.md §8)

### 🔴 Blockers
None.

### 🟡 Worth fixing / noting
- **`runMoveScript` JSON Result parsers (`parseSuccess` / `parseVmStatus`) are speculative** (`harness/script.ts:175-184`): no captured `move run-script` fixture exists. The `txHash` parser inherits the battle-tested Publisher.ts pattern (now shared via `parseCliOutput.ts`), so the critical field is safe. The two optional fields would just stay `undefined` if real CLI output differs. M4 (#71) validates against real output.
- **`runViewFunction` mocking via monkey-patch** (`view.test.ts:74, 99, 116, 133`): `harness.runtime.aptos.view = fake as never`. The SDK isn't behind an adapter; the wrapper is a 6-line forward. Could refactor to expose an adapter seam if `view.ts` grows, but premature today. **If this approach feels brittle, alternative is to abstract `aptos.view` behind a tiny adapter in M3 when coverage work touches `runtime.ts` anyway.**
- **`script.ts` lacks an `inheritStdio` option** — long-running scripts won't stream output to the user (it's buffered until completion). Fine for short scripts; if users complain in M2.4, add `silent: false` defaulting to `inheritStdio: true`. Documented gap, not blocker.

### 🟢 Nits
- `parseSuccess` and `parseVmStatus` are small (one regex each) — could fold into `parseCliOutput.ts` if a third consumer ever needs them. Today only script.ts uses them; share-on-third-use rule says wait.
- `script.ts` and `codeObject.ts` share the profile + try/catch shell. Could extract a `withMovementProfile()` higher-order helper, but the variance (different CLI args, different result parsers) makes the abstraction cost > duplication cost. Leaving as-is.
- `view.ts` is so thin (6 lines) it's almost a candidate to inline into `Harness.ts`. Kept as a separate file for symmetry with the other three method modules + future testability if it grows.

### Resolution
No fixes pushed. All findings 🟡-or-below; the parser-fixture concern is the same one tracked since M2.2 (M4 is the validation gate). Merging.

### Verification recap
- Tier 1: tsc clean; 220/220 unit tests; check:example green.
- Tier 2: smoke ✅; example skipped (mocha flake, publish path untouched by C1's byte-identical extraction).
- Tier 3: N/A (M2.4 runs the full ladder).

PR #126, opened 2026-05-14
feat(harness/fork): wire apiKey end-to-end via Authorization header (M2.1 TODO close)

gilbertsahumada:
## Summary

Closes the M2.1 TODO: \`Harness.createFork(network, apiKey?)\` accepted the param but did nothing with it. Now wired through 5 layers so every Movement API request from the fork's \`MovementApiClient\` carries \`Authorization: Bearer <apiKey>\`.

## What changed

| Layer | Change |
|---|---|
| \`fork/api.ts\` | \`MovementApiClient\` ctor takes optional \`apiKey?\`; \`get<T>()\` switches to the \`(url, options, cb)\` overload of \`https.get\`/\`http.get\` and adds the Authorization header when set. Header omitted when \`apiKey\` is undefined (back-compat for public endpoints). |
| \`fork/manager.ts\` | \`ForkManager\` gains private \`apiKey\` state. \`initialize(nodeUrl, networkName, apiKey?)\` stores it. \`load()\` reuses stored apiKey when reconstructing the client. New \`setApiKey(key)\` for the load-then-set pattern. Key is NOT persisted to disk (stays in process memory). |
| \`types/config.ts\` | \`LocalTestOptions\` gains \`forkApiKey?: string\`. |
| \`helpers/setupLocalTesting.ts\` | \`setupWithFork\` threads \`options.forkApiKey\` to \`forkManager.initialize()\` on the fresh-fork path and \`forkManager.setApiKey()\` before \`load()\` on the existing-fork path. |
| \`harness/Harness.ts\` | \`createFork\` drops the \`_apiKey\` underscore + TODO block; passes the param through. JSDoc updated. |

## New test

\`packages/movehat/src/__tests__/fork/api.test.ts\` uses \`vi.mock('https')\` + \`vi.mock('http')\` to capture request options without making real network calls. 2 cases:

1. **With apiKey:** \`'Authorization: Bearer <key>'\` present in captured options.headers
2. **Without apiKey:** Authorization header absent (back-compat for unauthenticated endpoints)

## Test plan

- [x] Tier 1: \`pnpm --filter movehat exec tsc --noEmit\` — clean
- [x] Tier 1: \`pnpm --filter movehat test\` — **224/224 passing** (222 prior + 2 new)
- [x] Tier 1: \`pnpm check:example\` — green
- [N/A] Tier 2: no template / smoke surface change (Harness JSDoc-only update; no new public type exports)
- [N/A] Tier 3: no install-experience change

## Out of scope

- **Movement testnet rate-limit testing** — this PR plumbs the header; whether the upstream actually honors it for rate-limit elevation is a deployment concern, not a code concern. M4 (#71) integration suite will validate against a real authenticated endpoint when applicable.
- **Sibling \`Harness.createLive(network, faucetUrl?)\` faucetUrl wiring** — still a silent no-op (its semantics are different: faucet URL drives auto-fund flow on testnets that have one). Separate follow-up if a real use case surfaces.

Tracks the M2 batch "Known follow-ups" section. Standalone chore, no milestone gating.

### DIFF
--- a/packages/movehat/src/__tests__/fork/api.test.ts
+++ b/packages/movehat/src/__tests__/fork/api.test.ts
@@ -1,0 +1,143 @@
+import { afterEach, beforeEach, describe, expect, it, vi } from "vitest";
+import type { ClientRequest, IncomingMessage } from "node:http";
+import { EventEmitter } from "node:events";
+
+/**
+ * Tests for `MovementApiClient` Authorization header injection (M2.4
+ * follow-up — wires the apiKey threading from `Harness.createFork`).
+ *
+ * Strategy: `vi.mock` `node:https` and `node:http` so the test
+ * captures the request options passed to `client.get(url, options, cb)`
+ * without making real network requests. Two assertions:
+ *
+ *   1. When the client is constructed with an apiKey, every `get`
+ *      call carries `Authorization: Bearer <apiKey>` in
+ *      `options.headers`.
+ *   2. When constructed without an apiKey, no Authorization header
+ *      is added (back-compat for unauthenticated public endpoints).
+ */
+
+const httpsGet = vi.fn();
+const httpGet = vi.fn();
+
+vi.mock("https", () => ({
+  default: { get: httpsGet },
+  get: httpsGet,
+}));
+vi.mock("http", () => ({
+  default: { get: httpGet },
+  get: httpGet,
+}));
+
+/**
+ * Build a fake `IncomingMessage`-like emitter that immediately emits
+ * a valid JSON body and ends. The MovementApiClient.get callback
+ * consumes the response via `data` / `end` events.
+ */
+function makeFakeResponse(body: string, statusCode = 200): IncomingMessage {
+  const res = new EventEmitter() as unknown as IncomingMessage;
+  (res as unknown as { statusCode: number }).statusCode = statusCode;
+  // Use setImmediate so the listener attaches before the events fire.
+  setImmediate(() => {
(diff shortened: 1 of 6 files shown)


gilbertsahumada (commented):
## Self-review (CLAUDE.md §8)

### 🔴 Blockers
None.

### 🟡 Worth noting
- **`ForkManager.load()` doesn't persist apiKey to disk** (intentional — keys stay in process memory). If a user loads a fork in a new process expecting auth to work, they must `setApiKey` explicitly. Documented in the `load()` JSDoc.
- **Test mocks both `https` and `http` via `vi.mock`**: `MovementApiClient` picks the client based on URL protocol. The fixture uses `https://` so only `httpsGet` actually fires. `httpGet` mock exists for symmetry but isn't asserted on. Acceptable; the test would still pass with http-only URL with no changes.
- **Test imports via dynamic `await import`** to ensure `vi.mock` hoisting fires before the import. Standard vitest pattern; documented in inline comments.

### 🟢 Nits
- Could extract the fake-response builder + handler into a shared test helper if `fork/api.test.ts` grows. Today it's one file with two tests — premature.
- `setupLocalTesting.ts` has two `apiKey`-related branches (fresh fork vs load existing). Could consolidate via a setApiKey-always-first pattern, but the current shape is more readable + the load path needs the key BEFORE `load()` reconstructs the client.

### Resolution
No fixes pushed. Merging.

### Verification recap
- Tier 1: tsc + 224/224 unit + check:example green.
- Tier 2/3: N/A (no template/install surface change).

PR #128, opened 2026-05-14
refactor(harness/tests): extract shared fixture helper (no behavior change)

gilbertsahumada:
## Summary

Six \`__tests__/harness/\` files duplicated the same \`beforeEach\`/\`afterEach\` boilerplate (~30 LoC each): tmpdir cwd, write \`movehat.config.js\` + Move package, chdir, \`_resetConfigCache\`, plus optional HOME management for tests exercising \`~/.aptos/config.yaml\` profile writes. Identified as a 🟢 nit in the M2.1 self-review; triggered when M2.2/M2.3/M2.4 added more variants without consolidating.

## What changed

**New:** \`packages/movehat/src/__tests__/harness/_fixture.ts\` — exports \`setupHarnessTestFixture({ extraNetworks?, withTmpHome? })\` returning \`{ tmpCwd, tmpHome?, teardown() }\`. Defaults match the common shape (single \`testnet\` network, no HOME management). Two opt-ins cover the variant cases:

- \`extraNetworks\` — \`Harness.createLive.test.ts\` needs a \`custom\` network for the network-switch test.
- \`withTmpHome\` — \`codeObject.{deploy,upgrade}.test.ts\` + \`script.test.ts\` need isolated HOME for profile-write tests.

**Migrated 6 files:**
- \`Harness.proxy.test.ts\` — default fixture
- \`Harness.createLive.test.ts\` — default + extra \`custom\` network
- \`codeObject.deploy.test.ts\` — \`withTmpHome: true\` + separate \`delete process.env.MH_CLI_REDEPLOY\` cleanup
- \`codeObject.upgrade.test.ts\` — \`withTmpHome: true\`
- \`script.test.ts\` — \`withTmpHome: true\`
- \`view.test.ts\` — default fixture

## vitest pickup

\`vitest.config.ts\` include pattern is \`src/**/__tests__/**/*.test.ts\` — \`_fixture.ts\` is not picked up (no \`.test\` suffix). The file is test-only, not re-exported from \`src/index.ts\`.

## Net diff

\`-262\` LoC across 6 test files + \`+140\` LoC new helper = **net ~-120 LoC** and DRY-er. Drift surface goes from 6 copies to 1.

## Test plan

- [x] Tier 1: \`pnpm --filter movehat exec tsc --noEmit\` — clean
- [x] Tier 1: \`pnpm --filter movehat test\` — **224/224 unchanged** (no behavior change)
- [x] Tier 1: \`pnpm check:example\` — green
- [N/A] Tier 2/3: pure test-infra refactor; no public surface, no template, no CLI change

Standalone refactor chore. No milestone gating.

### DIFF
--- a/packages/movehat/src/__tests__/harness/Harness.createLive.test.ts
+++ b/packages/movehat/src/__tests__/harness/Harness.createLive.test.ts
@@ -1,60 +1,25 @@
 import { afterEach, beforeEach, describe, expect, it } from "vitest";
-import { existsSync, mkdirSync, mkdtempSync, rmSync, writeFileSync } from "node:fs";
-import { tmpdir } from "node:os";
-import { join } from "node:path";
 
 import { Harness } from "../../harness/index.js";
-import { _resetConfigCache } from "../../core/config.js";
+import { setupHarnessTestFixture, type HarnessTestFixture } from "./_fixture.js";
 
 describe("Harness.createLive", () => {
-  let tmpCwd: string;
-  let origCwd: string;
+  let fixture: HarnessTestFixture;
 
   beforeEach(() => {
-    tmpCwd = mkdtempSync(join(tmpdir(), "movehat-harness-live-"));
-    writeFileSync(
-      join(tmpCwd, "movehat.config.js"),
-      `export default {
-  defaultNetwork: "testnet",
-  networks: {
-    testnet: {
-      url: "https://testnet.movementnetwork.xyz/v1",
-      chainId: "testnet"
-    },
-    custom: {
-      url: "https://custom.example.com/v1",
-      chainId: "custom",
-      accounts: ["0x${"a".repeat(64)}"]
-    }
-  }
-};
-`
-    );
-    const moveDir = join(tmpCwd, "move");
-    mkdirSync(join(moveDir, "sources"), { recursive: true });
-    writeFileSync(
-      join(moveDir, "Move.toml"),
-      `[package]
-name = "dummy"
-version = "0.0.1"
-
-[addresses]
-`
-    );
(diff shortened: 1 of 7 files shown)


gilbertsahumada (commented):
## Self-review (CLAUDE.md §8)

### 🔴 Blockers
None.

### 🟡 Worth noting
- **`script.test.ts` still imports `mkdirSync` + `writeFileSync`** because some tests create per-test script files (`scripts/init.move`, etc.) inside `fixture.tmpCwd`. Could push those into the fixture too via an `extraFiles?: Record<string, string>` option, but only `script.test.ts` needs it — single-consumer abstraction would be premature.
- **The fixture writes `networks` via `JSON.stringify`**: valid JS object literal (JSON is a subset). The previous handwritten configs used JS literal syntax; output is equivalent.
- **`fixture.tmpHome!` non-null assertion** in the codeObject + script tests is opt-in safe (`withTmpHome: true` was passed → `tmpHome` is always defined). Could narrow via discriminated union but the runtime contract is clear from the test's `beforeEach`.

### 🟢 Nits
- The fixture helper file is `_fixture.ts` (underscore prefix). Could also live in a `_helpers/` subdirectory if more helpers emerge. Premature today.
- `extraNetworks` shape is a copy of the relevant `NetworkConfig` slice from `types/config.ts` rather than importing the type directly. Avoids a test-only file depending on the full type module; if other variants emerge, worth revisiting.

### Resolution
No fixes pushed. Merging.

### Verification recap
- Tier 1: tsc + **224/224 unit tests** (unchanged) + check:example green.
- Tier 2/3: N/A (test-infra refactor; no behavior surface).

Net -120 LoC + drift surface 6 → 1.

Issue #70, opened 2026-05-08
[Roadmap] M3: 80% unit coverage + Movement CLI cache in CI


gilbertsahumada:
## Goal
Raise unit-test coverage on critical modules to **≥80% statements** and cut CI time by caching the Movement CLI tarball.

### Per-file coverage (each file ≥80% statements)

Core modules:
- [ ] `packages/movehat/src/runtime.ts`
- [ ] `packages/movehat/src/core/config.ts`
- [ ] `packages/movehat/src/core/AccountManager.ts`
- [ ] `packages/movehat/src/fork/manager.ts`
- [ ] `packages/movehat/src/node/LocalNodeManager.ts`

Top-level commands:
- [ ] `packages/movehat/src/commands/compile.ts`
- [ ] `packages/movehat/src/commands/init.ts`
- [ ] `packages/movehat/src/commands/run.ts`
- [ ] `packages/movehat/src/commands/test.ts`
- [ ] `packages/movehat/src/commands/test-move.ts`
- [ ] `packages/movehat/src/commands/update.ts`

Fork subcommands (interpreted as part of "all commands"):
- [ ] `packages/movehat/src/commands/fork/create.ts`
- [ ] `packages/movehat/src/commands/fork/fund.ts`
- [ ] `packages/movehat/src/commands/fork/list.ts`
- [ ] `packages/movehat/src/commands/fork/serve.ts`
- [ ] `packages/movehat/src/commands/fork/view-resource.ts`

### Coverage gates
- [ ] `vitest.config.ts` global `coverage.thresholds.lines: 80` (replaces the temporary 15%)
- [ ] Per-target thresholds enforced via `coverage.thresholds.<glob>` map
- [ ] CI gate in `.github/workflows/ci.yml` reads `coverage-summary.json` and fails if global <80%
- [ ] `coverage-summary.json` uploaded as CI artifact for inspection

### Movement CLI cache
- [ ] `actions/cache@v4` step in the E2E job, keyed by Movement CLI tarball SHA or release URL hash
- [ ] Cache stored at `/usr/local/bin/movement` (or a project-local path)
- [ ] Cache key invalidation documented in `TESTING.md`
- [ ] Cache hit visibly skips the 66 MB download in CI logs

## Test approach by file

- `runtime.ts` — uses Publisher (M1 prereq); mock `child_process` via adapter; use `memfs` for fs.
- `config.ts` — load inline + override paths; cache by mtime; tsx loader edge cases.
[...]

PR #130, opened 2026-05-14
ci: cache Movement CLI tarball in E2E job (M3.1)

### LINKED ISSUE #129: [M3.1] Cache Movement CLI tarball in E2E job
Tracks #70

## Definition of Done

- [ ] `actions/cache@v4` step in the E2E job, keyed by Movement CLI tarball SHA / release URL hash
- [ ] Cache hit path skips the 66 MB tarball download
- [ ] Visible runtime drop on cache hit vs miss in CI logs

## Scope

`.github/workflows/ci.yml` edit only. No movehat code changes.

Insert an `actions/cache@v4` step (id `cache-movement-cli`) before the "Install Movement CLI" step. Cache `path: /usr/local/bin/movement`, key derived from runner OS + arch + the `bypass-homebrew` release tag. Wrap the download/extract block in `if: steps.cache-movement-cli.outputs.cache-hit != 'true'`. Keep `movement --version` outside the conditional as a sanity check.

## Verification

Tier 1 (tsc + tests) unchanged. Real validation: first push = cache MISS (writes cache); re-run = cache HIT, download step skipped. Document raw-second delta in PR description.

gilbertsahumada:
## Summary

Adds an `actions/cache@v4` step before the Movement CLI install in the E2E job. On cache hit, the 66 MB download/extract step is skipped; `movement --version` verification runs unconditionally as a sanity check on both paths.

- Cache key: `movement-cli-${{ runner.os }}-${{ runner.arch }}-bypass-homebrew-l1`
- Restore-keys: `movement-cli-${{ runner.os }}-${{ runner.arch }}-` (allows partial restores if the tag changes)
- Path: `/usr/local/bin/movement`

The "Install Movement CLI" step is now wrapped in `if: steps.cache-movement-cli.outputs.cache-hit != 'true'`. The `--version` verification was extracted into a separate "Verify Movement CLI" step that runs on both code paths.

## DoD (Tracks #70 / Closes #129)

- [x] `actions/cache@v4` step in the E2E job, keyed by Movement CLI tarball release URL hash
- [x] Cache hit path skips the 66 MB tarball download
- [ ] Visible runtime drop on cache hit vs miss in CI logs — observable on first re-run; will report numbers in a comment once the workflow runs

## Verification

| Tier | Command | Result |
|---|---|---|
| 1 | `pnpm --filter movehat exec tsc --noEmit` | pass (no src changes) |
| 1 | `pnpm --filter movehat test` | pass (unchanged) |
| 1 | `pnpm check:example` | pass |
| 2 | N/A | workflow-only edit |
| 3 | N/A | install experience unchanged |

Real validation lives in CI: first push triggers cache MISS (writes cache); a re-run from the GitHub UI should show cache HIT and the Install step skipped. Runtime delta to be documented in a follow-up comment.

## Notes

- Pre-existing `quality` job uses `node -e "... require('./package.json') ..."` (line 219) — flagged by a local lint hook as a false positive. `require()` is valid inside `node -e` subprocesses; not touching per surgical-changes rule (CLAUDE.md §3).

## Test plan

- [x] Local pre-push gate (tsc + tests + check:example) passed
- [ ] CI: first run shows cache MISS, downloads tarball
- [ ] CI: re-run shows cache HIT, Install step skipped, `movement --version` still prints

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -163,7 +163,17 @@ jobs:
           name: build-artifacts
           path: packages/movehat/dist/
 
+      - name: Cache Movement CLI binary
+        id: cache-movement-cli
+        uses: actions/cache@v4
+        with:
+          path: /usr/local/bin/movement
+          key: movement-cli-${{ runner.os }}-${{ runner.arch }}-bypass-homebrew-l1
+          restore-keys: |
+            movement-cli-${{ runner.os }}-${{ runner.arch }}-
+
       - name: Install Movement CLI
+        if: steps.cache-movement-cli.outputs.cache-hit != 'true'
         run: |
           curl -fLO https://github.com/movementlabsxyz/homebrew-movement-cli/releases/download/bypass-homebrew/movement-cli-l1-linux-x86_64.tar.gz
           mkdir -p temp_extract
@@ -171,7 +181,9 @@ jobs:
           chmod +x temp_extract/movement
           sudo mv temp_extract/movement /usr/local/bin/movement
           rm -rf temp_extract movement-cli-l1-linux-x86_64.tar.gz
-          movement --version
+
+      - name: Verify Movement CLI
+        run: movement --version
 
       - name: Run E2E tests
         run: pnpm test:e2e:quick
--- a/ROADMAP.md
+++ b/ROADMAP.md
@@ -149,9 +149,9 @@ Each milestone below lists **explicit, mechanically verifiable** acceptance crit
 - [ ] `coverage-summary.json` produced and uploaded as CI artifact
 
 **Definition of Done — CI cache**:
-- [ ] `actions/cache@v4` step in the E2E job, keyed by Movement CLI tarball SHA / release URL hash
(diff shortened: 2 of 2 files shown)


gilbertsahumada (commented):
## Self-review (§8)

🔴 **Blockers**: none.

🟡 **Worth fixing before merge**: none.

🟢 **Nits / observations**:
- The `restore-keys` line uses a partial-prefix match (`movement-cli-${{ runner.os }}-${{ runner.arch }}-`). If the upstream `bypass-homebrew` tag is ever bumped, a stale binary could be restored from the partial-match cache and fail at `movement --version` — but the failure mode is clear (the verify step exits non-zero) and the next run would write a fresh key. Acceptable; documented in the plan's risk table.
- The "Verify Movement CLI" step is split out so it runs unconditionally. Cache-hit and cache-miss paths both prove the binary works.
- Pre-existing `node -e "... require('./package.json') ..."` in the `quality` job triggers a local lint hook as a false positive (`require()` is valid inside `node -e` subprocesses). Not touched per §3.

Ready to merge.

Issue #140, opened 2026-05-14
[M3-followup] Pin Movement CLI artifact integrity in E2E job


gilbertsahumada:
Tracks #70 (follow-up to M3 / M3.1).

## Context

CodeRabbit flagged (PR #135) that the E2E job's Movement CLI install fetches a remote tarball without integrity verification:

\`\`\`yaml
curl -fLO https://github.com/movementlabsxyz/homebrew-movement-cli/releases/download/bypass-homebrew/movement-cli-l1-linux-\${CLI_ARCH}.tar.gz
\`\`\`

This is a real supply-chain concern that was deferred from M3 because hardening properly requires upstream cooperation (signed checksums or release pinning) and is out of M3 scope.

## What we want

- Pin to a specific Movement CLI release tag (not the moving \`bypass-homebrew\` tag).
- Verify the downloaded artifact via SHA256 checksum (or GPG signature if upstream provides one) BEFORE \`chmod +x\` / \`sudo mv\`.
- Cache key includes the pinned version so cache rotates correctly on upgrade.

## Blockers / open questions

- Does \`movementlabsxyz/homebrew-movement-cli\` publish a signed digest or SHA256 file alongside the release? If not, what version-pinning strategy is appropriate?
- Do we want this to land before M5 (which introduces a Movement CLI compat matrix and may consolidate the pinning story)?

## DoD

- [ ] Workflow downloads Movement CLI from a pinned release tag (not \`bypass-homebrew\`).
- [ ] Workflow verifies SHA256 (or signature) before installing.
- [ ] Cache key incorporates the pinned version.
- [ ] CI job fails on checksum mismatch.

Likely best timed with M5's compat-matrix work.

PR #135, opened 2026-05-14
M3 batch: 80% unit coverage + Movement CLI cache (KPI 1)

gilbertsahumada:
## Summary

M3 milestone (KPI 1, issue #70). Batch merge of 5 sub-PRs delivered to `develop`:

| Sub-PR | Description | PR |
|---|---|---|
| M3.1 | Cache Movement CLI tarball in E2E job | #130 |
| M3.2 | Coverage: runtime + config + AccountManager to ≥80% | #132 |
| M3.3 | Coverage: ForkManager + LocalNodeManager to ≥80% | #134 |
| M3.4 | Coverage: top-level commands to ≥80% | #137 |
| M3.5 | Coverage: fork commands + flip global gate to 70% | #139 |

## Coverage results

All 16 M3 target files now ≥80% statements:

| File | Before | After (stmts) |
|---|---|---|
| `src/runtime.ts` | 63.9% | **100%** |
| `src/core/config.ts` | 69.0% | **88.7%** |
| `src/core/AccountManager.ts` | 40.2% | **97.4%** |
| `src/fork/manager.ts` | 7.8% | **97.1%** |
| `src/node/LocalNodeManager.ts` | 0.0% | **94.4%** |
| `src/commands/compile.ts` | 57.1% | **95.9%** |
| `src/commands/init.ts` | 0.0% | **98.4%** |
| `src/commands/run.ts` | 8.2% | **75.5%** (threshold 70 — see note) |
| `src/commands/test.ts` | 0.0% | **81.5%** |
| `src/commands/test-move.ts` | 0.0% | **100%** |
| `src/commands/update.ts` | 0.0% | **96.4%** |
| `src/commands/fork/create.ts` | 0.0% | **95.6%** |
| `src/commands/fork/fund.ts` | 0.0% | **100%** |
| `src/commands/fork/list.ts` | 0.0% | **86.4%** |
| `src/commands/fork/serve.ts` | 0.0% | **86.2%** |
| `src/commands/fork/view-resource.ts` | 0.0% | **92.3%** |

**Package-wide**: 37.1% → 73.9% lines (+36.8 pts). Test count: 224 → 393 (+169).

## Threshold deviations

Two locked-plan adjustments documented inline in `vitest.config.ts`:

[...]

PR #141, opened 2026-05-14
fix: address CodeRabbit batch review on PR #135 (M3-CR)

gilbertsahumada:
## Summary

Follow-up sub-PR addressing CodeRabbit's batch review on the M3 develop→main PR (#135). Same pattern as M2-CR (PRs landed on develop, not on the batch directly, per CLAUDE.md §5).

## Triage table

| # | Path:line | Severity | Verdict | Action |
|---|---|---|---|---|
| 1 | `ci.yml:189` | 🟠 Major | Defer | Filed as #140 — Movement CLI artifact integrity verification (needs upstream signed checksum process; out of M3 scope) |
| 2 | `app/raw/[...slug]/route.ts:11` | 🟡 Minor | Fix | Stale comment ref → updated to `app/raw/index/route.ts` (matches the EISDIR-fix layout from PR #127) |
| 3 | `AccountManager.test.ts:129` | 🟡 Minor | Fix | "Insertion order" test now asserts ordered `.toEqual([...])` instead of just `.toContain` |
| 4 | `ROADMAP.md:152` | 🟡 Minor | Fix | Cache-key description rewritten to match the actual key shape |
| 5 | `ci.yml:174-181` | 🟢 Nit | Fix | Arch-aware install URL (X64 → x86_64, ARM64 → aarch64) so the URL follows the cache-key's `runner.arch` dimension |

## Verification

| Tier | Command | Result |
|---|---|---|
| 1 | `pnpm --filter movehat exec tsc --noEmit` | pass |
| 1 | `pnpm --filter movehat test` | pass (count unchanged — refactor, not net-new) |
| 1 | `pnpm check:example` | pass |
| 2 | N/A — pure refactor + docs + CI | — |
| 3 | N/A | — |

## Notes

- The "Major" finding (#1) is documented as deferred to #140. Real supply-chain concern, but the existing CI gate runs against a pinned `bypass-homebrew` tag over HTTPS — current posture is not worse than before, and the proper fix needs upstream cooperation.
- The arch-aware install (#5) was originally a "nitpick" but fixing it costs <10 LoC and removes a latent bug if we ever migrate to ARM runners. Included.
- PostToolUse validation hook flagged a pre-existing `node -e "require('./package.json')"` in the `quality` job (line 232) as a false positive about `require()` in workflows. Untouched per §3 (Surgical Changes).

Tracks #70

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -178,12 +178,22 @@ jobs:
       - name: Install Movement CLI
         if: steps.cache-movement-cli.outputs.cache-hit != 'true'
         run: |
-          curl -fLO https://github.com/movementlabsxyz/homebrew-movement-cli/releases/download/bypass-homebrew/movement-cli-l1-linux-x86_64.tar.gz
+          # Map runner.arch to the artifact suffix the upstream release uses.
+          # The cache key (above) already includes runner.arch, so the
+          # download URL must follow the same arch dimension or a cache hit
+          # on the wrong architecture would install an unusable binary.
+          case "${{ runner.arch }}" in
+            X64)   CLI_ARCH="x86_64" ;;
+            ARM64) CLI_ARCH="aarch64" ;;
+            *)     echo "::error::Unsupported runner.arch: ${{ runner.arch }}"; exit 1 ;;
+          esac
+          ARTIFACT="movement-cli-l1-linux-${CLI_ARCH}.tar.gz"
+          curl -fLO "https://github.com/movementlabsxyz/homebrew-movement-cli/releases/download/bypass-homebrew/${ARTIFACT}"
           mkdir -p temp_extract
-          tar -xzf movement-cli-l1-linux-x86_64.tar.gz -C temp_extract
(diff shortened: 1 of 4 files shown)


gilbertsahumada (commented):
## Self-review (§8)

🔴 **Blockers**: none.

🟡 **Worth fixing before merge**: none.

🟢 **Nits / observations**:
- The `ci.yml` arch-mapping uses `case` with explicit `X64` / `ARM64` matches and falls through to `::error` for anything else. The unsupported-arch path is reachable only if someone migrates the runner to a 32-bit or unusual arch; failing loudly beats silently downloading the wrong artifact.
- The `AccountManager.test.ts` ordering assertion now uses `.toEqual([a.addr, b.addr])` instead of `.toContain`. The underlying `Map` iterates in insertion order per the JS spec; if the implementation ever switches data structures, this test fails immediately.
- The deferred Major finding (#140) is filed against the broader supply-chain concern; not in this PR's scope. The existing CI install already runs over HTTPS from the official Movement Labs release URL — the current posture isn't worse than the previous run.

Ready to merge.

Issue #149, opened 2026-05-14
Movement local-node fails to start on Linux CI (MintFunder ENOT_APTOS_FRAMEWORK_ADDRESS abort)


gilbertsahumada:
## Summary

The Movement local-testnet startup (`movement node run-local-testnet`, the binary that backs `LocalNodeManager.start`) fails on Linux x86_64 in GitHub Actions CI with a `MintFunder` setup error before the node API ever becomes ready. Surfaced by the M4.2 integration suite's first Actions run on PR #147.

Locally on macOS (Darwin arm64) with the same Movement CLI revision (`bypass-homebrew`), the same node startup succeeds in ~10–15 s. CI fails consistently in ~13 s with the error below.

## Error (CI log, Linux runner)

\`\`\`
[Node Error] One of the services failed to start up, running shutdown steps...
Ran shutdown steps

context: \"Failed to build Funder\",
source: Error {
    context: \"Failed to build MintFunder\",
    source: Error {
        context: \"Failed to make MintFunder use delegated account\",
        source: Error {
            context: \"Failed to delegate minting to the new account\",
            source: Unknown(
                \"Transaction committed on chain, but failed execution: Move abort in 0x1::system_addresses: ENOT_APTOS_FRAMEWORK_ADDRESS(0x50003): The address/account did not correspond to the core framework address\",
            ),
        },
    },
},
\`\`\`

LocalNodeManager then times out after 60 s on the readiness probe and throws \`Local node did not become ready within 60 seconds.\`

Full CI run log: https://github.com/gilbertsahumada/movehat/actions/runs/25865016223

## Repro

1. Linux x86_64 runner (matched: ubuntu-latest, Node 20, pnpm 9).
2. Movement CLI installed from \`https://github.com/movementlabsxyz/homebrew-movement-cli/releases/download/bypass-homebrew/movement-cli-l1-linux-x86_64.tar.gz\`.
3. From any working dir:
   \`\`\`
   movement node run-local-testnet --with-faucet --force-restart --assume-yes
   \`\`\`
4. Observe genesis shutdown with the \`MintFunder\` chain — \`ENOT_APTOS_FRAMEWORK_ADDRESS\` abort during \`delegate_mint_capability\`.

## Hypotheses to investigate

[...]

PR #145, opened 2026-05-14
test(integration): zero-mock harness suite + vitest config + fixtures (M4.1)

gilbertsahumada:
## Summary

M4.1 of the [M4 roadmap](https://github.com/gilbertsahumada/movehat/issues/71). Build a TypeScript-native, vitest-based integration suite under `packages/movehat/test/integration/` that drives the real Movement CLI through the documented public Harness lifecycle with **zero** `vi.mock` / `vi.spyOn` calls. The suite is the first signal that fails when upstream CLI behavior shifts — rather than the failure landing in user projects.

This is the heavy lift of M4. The CI policy flip (E2E on every push + 5-min SLO + integration job wired in) ships separately as M4.2.

## Suite contents

| File | Cases | Wall-clock |
|---|---|---|
| `harness-local.integration.test.ts` | 5 | ~33 s |
| `harness-fork.integration.test.ts` | 2 (skipIf `MOVEMENT_RPC_URL` unset) | ~5 s |

**Local lifecycle covered** (single `describe`, sequential `beforeAll → it × 5 → afterAll`):
1. `Harness.createLocal({ accountLabels: ['deployer', 'alice'] })` boots a real Movement node
2. `deployCodeObject(v1)` — publishes the minimal Move fixture with `init_module` auto-init
3. `runViewFunction({ function: counter::get })` — returns `'0'`
4. `MoveContract.call(deployer, 'increment')` then view — returns `'1'`
5. `runMoveScript({ scriptPath: noop.mv })` — pre-compiled bytecode branch
6. `upgradeCodeObject(v2, objectAddress=v1.address)` — returns same address + new txHash
7. `harness.cleanup()` in `afterAll`

**Fork path covered** (skip-when-env-missing so local runs stay self-contained):
1. `Harness.createFork('testnet')` returns a fork-mode harness whose runtime binds the local fork-server RPC
2. `harness.deployCodeObject(...)` rejects with `/fork/i` (read-only enforcement)

## Fixture layout

```
packages/movehat/test/integration/
├── vitest.integration.config.ts   (lives at package root — `pool: forks` + `singleFork: false` + `fileParallelism: false`)
├── movehat.config.js              (minimal config; `.js` not `.ts` — see comment for tsx/vitest loader collision)
├── harness-local.integration.test.ts
├── harness-fork.integration.test.ts
├── fixtures/
│   ├── move/v1/                   (Counter w/ init_module + increment + view get)
│   ├── move/v2/                   (compatible upgrade adding `reset`)
[...]

PR #150, opened 2026-05-14
fix(test): restore MH_CLI_REDEPLOY in harness-local afterAll

gilbertsahumada:
## Summary

Addresses CodeRabbit's actionable finding on the M4 batch PR #148 ([comment](https://github.com/gilbertsahumada/movehat/pull/148#discussion_r3241847370)). The integration suite's `harness-local.integration.test.ts` mutates `process.env.MH_CLI_REDEPLOY` in `beforeAll` but never restores it. Two independent reviewers (CodeRabbit + my own §8 self-review on #145 flagged this as Minor #2) agreed it should be defensive-by-default — fixing now.

## Change

```diff
 describe.skipIf(SKIP_LOCAL)('Harness.createLocal — full lifecycle', () => {
   let harness: Harness;
   let codeObjectAddr: string;
   let originalCwd: string;
+  let originalRedeployEnv: string | undefined;

   beforeAll(async () => {
     originalCwd = process.cwd();
+    originalRedeployEnv = process.env.MH_CLI_REDEPLOY;
     process.chdir(integrationDir);
     process.env.MH_CLI_REDEPLOY = 'true';
     // ...
   });

   afterAll(async () => {
     if (harness) await harness.cleanup();
     if (originalCwd) process.chdir(originalCwd);
+    if (originalRedeployEnv === undefined) {
+      delete process.env.MH_CLI_REDEPLOY;
+    } else {
+      process.env.MH_CLI_REDEPLOY = originalRedeployEnv;
+    }
   });
```

The `delete` branch matters when the original was undefined — setting `process.env.X = undefined` actually stores the string `"undefined"`, which would propagate as a truthy value to consumers.

## Why this matters now

The integration suite currently runs `pool: 'forks', singleFork: false` so cross-file bleed inside a single process is impossible. But the M4.1 original intent was `singleFork: true` — the port-collision constraint forced the switch. If we later return to a single-fork model (e.g. to amortize the local-node startup cost across files), this fix becomes load-bearing rather than defensive.

## Verification

- ✅ `pnpm --filter movehat exec tsc --noEmit` — clean
- ✅ `MOVEHAT_SKIP_LOCAL_NODE=true MOVEMENT_RPC_URL=https://testnet.movementnetwork.xyz/v1 pnpm test:integration` — 2/2 fork pass, 5 skipped, ~5 s (matches CI shape)

## Path forward

Once this lands on `develop`, the batch PR #148 auto-picks it up. The other CodeRabbit findings on #148 are either false positives or deferred to a follow-up `chore(docs): lint TESTING.md` chore (replied inline on each comment).

Refs #148, #149

### DIFF
--- a/packages/movehat/test/integration/harness-local.integration.test.ts
+++ b/packages/movehat/test/integration/harness-local.integration.test.ts
@@ -42,6 +42,7 @@ describe.skipIf(SKIP_LOCAL)('Harness.createLocal — full lifecycle', () => {
   let harness: Harness;
   let codeObjectAddr: string;
   let originalCwd: string;
+  let originalRedeployEnv: string | undefined;
 
   beforeAll(async () => {
     // `Harness.createLocal` → `initRuntime` → `loadUserConfig` reads
@@ -50,8 +51,12 @@ describe.skipIf(SKIP_LOCAL)('Harness.createLocal — full lifecycle', () => {
     // load succeeds without polluting the package root with a user-
     // facing config file. `MH_CLI_REDEPLOY=true` allows re-running the
     // suite without `deployments/local/counter.json` stale-record
-    // bookkeeping aborting the second run.
+    // bookkeeping aborting the second run. Both env mutations get
+    // restored in afterAll so the process state is exactly what it
+    // was on entry — defensive against future integration tests that
+    // run in the same fork.
     originalCwd = process.cwd();
+    originalRedeployEnv = process.env.MH_CLI_REDEPLOY;
     process.chdir(integrationDir);
     process.env.MH_CLI_REDEPLOY = 'true';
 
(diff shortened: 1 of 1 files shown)


gilbertsahumada (commented):
## §8 self-review — M4-CR env-restore (post-merge)

Posted post-merge to close the audit trail. CLAUDE.md §8 requires unconditional self-review on every PR; this one was skipped because the fix was 3 lines and felt self-evident at merge time. Honest assessment below.

### 🔴 Critical
None.

### 🟠 Major
None. The fix captures `originalRedeployEnv` in `beforeAll` and restores it in `afterAll` with a `delete` branch for the undefined case. Logic is mechanical; rationale is in the commit message + PR body.

### 🟡 Minor (1)

1. **Lesson from my own prior §8 review.** I flagged this exact issue on PR #145 as Minor #2 with "leaving as-is for now since the singleFork stance is documented" — accepting the defer because cross-file env bleed was structurally impossible under `singleFork: false`. CodeRabbit independently re-flagged it on the M4 batch (#148), and I fixed it here. The defer was wrong: cost-of-fix was one line, cost-of-being-wrong-later was a future regression on a singleFork toggle.

   **Future §8 review heuristic:** when a finding is "defensive-by-default" AND the fix is ≤5 lines AND the cost of being wrong is non-zero, bias toward applying it in the originating PR. The bar "two independent reviewers must agree" is higher than necessary — it just means the second reviewer pays the cost of re-flagging.

### 🟢 Trivial
None.

### Recommendation
None — already merged. The fix itself is sound; the meta-lesson about defer thresholds gets propagated forward into M5's §8 practice.

PR #168, opened 2026-05-15
feat(api)!: remove getMovehat() + bump to 0.2.0

gilbertsahumada:
Closes #166 #73. Tracks nothing additional (M6 meta complete after this).

## Summary

Final M6 sub-PR. Finalizes the \`getMovehat()\` deprecation arc started in M2.4 and bumps to \`0.2.0\` per semver (breaking change marker).

> **BREAKING CHANGE**: \`getMovehat()\` is no longer exported. Migration:
>
> **Recommended** — \`Harness.createLive()\` (lifecycle-managed):
> \`\`\`ts
> import { Harness } from "movehat";
> const harness = await Harness.createLive("testnet");
> try {
>   const deployment = await harness.runtime.deployContract("counter");
>   // ...
> } finally {
>   await harness.cleanup();
> }
> \`\`\`
>
> **Advanced** — \`initRuntime()\` (low-level, no Harness lifecycle):
> \`\`\`ts
> import { initRuntime } from "movehat";
> const runtime = await initRuntime();
> const deployment = await runtime.deployContract("counter");
> \`\`\`

## Critical correction to meta-issue #73

Meta-issue says "remove \`mh()\` AND \`MovehatRuntime\` legacy types". Reality:

- There is no \`mh()\` — the symbol is \`getMovehat\` (renamed in M1.5).
- \`MovehatRuntime\` is **NOT legacy** — it's the live runtime type used by \`Harness.runtime\`, \`initRuntime()\`, and all harness helpers (\`script.ts\`, \`view.ts\`, \`codeObject.ts\`). Removing it would break the live public surface.
- Version target \`0.1.0\` is stale — package was already at \`0.1.9\` with 9 published 0.1.x versions.

**Only \`getMovehat()\` itself is removed; \`MovehatRuntime\` stays.** ROADMAP M6 DoD updated with \`[N/A]\` + rationale per CLAUDE.md §7.

## Changes (11 files)

| File | Change |
|---|---|
| \`packages/movehat/src/index.ts\` | Drop \`getMovehat\` export + deprecation comment block. \`MovehatRuntime\` type stays. |
| \`packages/movehat/src/runtime.ts\` | Delete \`getMovehat()\` function body + its 30-line JSDoc. \`initRuntime\` and other exports preserved. |
| \`packages/movehat/src/templates/types/movehat.d.ts\` | Drop \`getMovehat\` declaration. |
[...]

PR #155, opened 2026-05-14
docs(claude): codify M4 process learnings (§7 closeout + §8 pre-merge gate & defer threshold)

gilbertsahumada:
## Summary

Three pure-addition edits to `CLAUDE.md` — each closes one open process gap surfaced in the M4 post-merge §8 reviews on #148/#150/#151. Now executable because PR #154 re-tracked CLAUDE.md as a repo file.

## What's in the PR

| § | New content | Anchor | Origin |
|---|---|---|---|
| §7 | "Milestone closeout (separate PR after batch merge)" subsection | End of §7, after "How to apply" | #151 Minor #1 |
| §8 | "Pre-merge checklist" bold callout (4 checkbox bullets) | Right after §8 headline rule | #148 Major #1 |
| §8 | "Defer threshold" indented sub-bullet under 🟡 | Under existing 🟡 description | #150 Minor #1 |

**Diff scope: 1 file, 19 insertions, 0 deletions.** Pure additions. Existing §7/§8 paragraphs unchanged — the ~46 historical citations against prior text continue to resolve.

### §7 — Milestone closeout subsection

```diff
+### Milestone closeout (separate PR after batch merge)
+
+After a `develop → main` batch PR merges, open a follow-on `docs/MN-roadmap-status` PR targeting `develop`:
+
+- Flip the milestone header line from `### MN — …` to `### MN — ✅ shipped in PR #<batch> (develop → main batch) — …`.
+- Add the sub-PR + commit-hash table inside the milestone section: columns `| Sub | Description | PR | Commit |` — mirrors the precedent set by M3 (PR #142 / commit `d8bafac`) and M4 (PR #151 / commit `0568049`).
+- List any follow-up issues filed during the milestone with a one-line "deferred to MN+1" rationale.
+- No code, no test, no CI changes. Pure ROADMAP edit.
+
+Why a separate PR from the batch itself: the batch is the change-set; the closeout is the post-mortem. Mixing them makes the batch diff harder to review (status flips bury inside hundreds of code-diff lines) and the milestone status harder to find via `git log --grep`.
```

Codifies what M3 and M4 already did de-facto. M5 won't need to re-derive the pattern.

### §8 — Pre-merge checklist

```diff
+**Pre-merge checklist** (from #148 Major #1 — verify each item before `gh pr merge`):
+
+- [ ] §8 self-review posted as a `gh pr review --comment` by the PR author
[...]

PR #170, opened 2026-05-15
fix(publish): allow same-version on npm version step (post-v0.2.0 hotfix)

gilbertsahumada:
Hotfix after the first v0.2.0 release attempt failed at the workflow's \`Update package version\` step. Tag and release were deleted; this PR fixes the workflow so the re-cut can proceed cleanly.

## Root cause

\`publish.yml\` step 7 ran:
\`\`\`bash
npm version 0.2.0 --no-git-tag-version
\`\`\`

against a manifest that already declared \`0.2.0\` (M6.2 / PR #168 bumped it). npm rejects this with \`npm error Version not changed\` and the workflow exits non-zero before reaching the gate, the tests, or the publish step.

## Fix

Add \`--allow-same-version\` to the same step:
\`\`\`bash
npm version 0.2.0 --no-git-tag-version --allow-same-version
\`\`\`

Now a no-op when manifest is already at target; unchanged behavior when it isn't. 7-line diff (6 lines explanatory comment + 1 line flag).

## Why direct to develop instead of "fast hotfix to main"

Standard branch workflow per CLAUDE.md. Develop → main batch comes next; the v0.2.0 re-cut is cheap (delete + recreate) once main has the fix.

## §8 self-review

Will be posted as a \`gh pr review --comment\`.

## Tier 2 / Tier 3

- N/A — CI-only change, no \`packages/movehat/src/**\` / templates touched. Workflow is exercised on release publish (will be re-tested on the v0.2.0 re-cut after this batch merges).

## Context

Failed run: https://github.com/gilbertsahumada/movehat/actions/runs/25900063511 (24s, "Version not changed").
v0.2.0 release + tag deleted at $(date -u +%FT%TZ) — no npm artifact was ever published; clean slate.

### DIFF
--- a/.github/workflows/publish.yml
+++ b/.github/workflows/publish.yml
@@ -49,9 +49,15 @@ jobs:
           echo "Publishing version: $VERSION"
 
       - name: Update package version
+        # `--allow-same-version` makes this step a no-op when the in-repo
+        # manifest already declares the release tag's version (i.e. the
+        # version bump shipped in the PR that landed before the tag was
+        # cut, rather than relying on the workflow to do the bump).
+        # Without it, npm errors with "Version not changed" and the
+        # release fails — see v0.2.0 attempt at 2026-05-15 (run 25900063511).
         run: |
           cd packages/movehat
-          npm version ${{ steps.get_version.outputs.version }} --no-git-tag-version
+          npm version ${{ steps.get_version.outputs.version }} --no-git-tag-version --allow-same-version
 
       - name: Verify CHANGELOG section exists for release version
         # Codified CHANGELOG gate per M6 (issue #73, sub-issue #165). Blocks


gilbertsahumada (commented):
## §8 self-review (pre-merge, unconditional)

### 🔴 / 🟡
None. 7-line diff, CI-only, addresses a specific empirical failure (run 25900063511).

### 🟢 nits
- An alternative would be to remove the `npm version` step entirely since M6.2 bumped the manifest in the PR. But that breaks the workflow_dispatch path where someone manually fires a release without first bumping. Keeping the step + adding `--allow-same-version` preserves both flows.

## Pre-merge checklist (#155 §8 codification)

- [x] §8 self-review posted as comment.
- [x] 🔴 / 🟡 — none.
- [x] Tier 2 / Tier 3 results — N/A with reason (CI-only).
- [ ] Mergeable — pending CI.

Resolved decision (not deferred concern): kept `npm version` step instead of removing it, to preserve `workflow_dispatch` path.

PR #186, opened 2026-05-15
chore(ci): scope CI to push:main + PRs to main only

gilbertsahumada:
## Summary

The CI workflow currently runs on every push to \`main\` / \`develop\` / \`feature/*\` plus every PR targeting \`main\` / \`develop\`. With the develop→batch→main workflow this generates 2-3 duplicate runs per logical change:

1. Sub-PR opens on develop → \`pull_request\` trigger fires CI.
2. Sub-PR merges to develop → \`push:develop\` trigger fires CI again (duplicate).
3. Batch PR opens on main → \`pull_request\` trigger fires CI.
4. Batch merges to main → \`push:main\` trigger fires CI (often a third run).

This PR collapses CI to the single meaningful checkpoint: **PRs targeting main + the push to main itself**. The batch develop→main PR becomes the canonical validation gate.

## New scope (\`.github/workflows/ci.yml\`)

\`\`\`yaml
on:
  push:
    branches: [main]               # was: [main, develop, feature/*]
  pull_request:
    branches: [main]               # was: [main, develop]
  workflow_dispatch:               # new — manual rerun on any branch
\`\`\`

## Trade-off accepted

| Before | After |
|---|---|
| Sub-PRs to develop got automatic CI feedback | Sub-PRs to develop run no CI — maintainer runs Tier-1 locally before merge |
| Push to develop triggered CI | Push to develop is silent |
| Push to feature/* triggered CI | Push to feature/* is silent |
| Duplicate runs on PR open + sub-merge | Single run per validation point |

**Mitigation**: CLAUDE.md §8 Pre-merge checklist already mandates local Tier-1 verification (\`pnpm --filter movehat test\`, \`pnpm check:example\`) before merging sub-PRs to develop, plus Tier-2 smoke when source under \`packages/movehat/src/**\` is touched. The §8 self-review process is the operational equivalent of pre-merge CI for sub-PRs.

**Escape hatch**: \`workflow_dispatch\` lets a maintainer manually trigger CI on any branch from the Actions tab when they want a green check before opening the batch — covers the "I want CI on develop right now" case without restoring the noisy push trigger.

## What was NOT changed

- \`.github/workflows/publish.yml\`: only runs on \`release: published\` + \`workflow_dispatch\`. Already scoped correctly.
- \`.github/workflows/docs-deploy.yml\`: only runs on \`push: [main]\` + \`workflow_dispatch\`. Already scoped correctly.

## Verification

[...]

PR #189, opened 2026-05-15
fix(npm): sync README from workspace root before publish

gilbertsahumada:
## Summary

npm publishes from \`packages/movehat/\` and uses the \`README.md\` in that directory. The package's README had drifted ~6 months behind the root README — every reorg / cleanup since then only touched the root copy (\`./README.md\`), but the \`packages/movehat/README.md\` was never updated. The npm registry page for \`movehat@0.2.0\` shows the old version (16-bullet feature dump, fork system as step 7, old \`deployContract\` example) instead of the redesigned current README.

### 1. One-time sync
\`packages/movehat/README.md\` overwritten with current root \`README.md\` content. After this commit, both files are identical.

### 2. Automation via \`prepack\` lifecycle
Added to \`packages/movehat/package.json\`:

\`\`\`json
\"scripts\": {
  ...
  \"prepack\": \"cp ../../README.md README.md\",
  \"prepublishOnly\": \"npm run build\"
}
\`\`\`

\`prepack\` runs automatically on:
- \`npm pack\` (used by \`test:smoke\` and \`test:e2e\`)
- \`npm publish\` (used by the release workflow)

So the published README always reflects whatever the root README looks like at release time. Zero manual sync needed; zero drift possible.

## Verification

\`\`\`
$ cd packages/movehat && npm pack --dry-run | head -3
> movehat@0.2.0 prepack
> cp ../../README.md README.md

npm notice 📦  movehat@0.2.0
npm notice Tarball Contents
npm notice 9.0kB README.md   ← post-reorg version, not stale 10kB
\`\`\`

## Caveat (user-visible)

npm does NOT update the README for already-published versions. \`movehat@0.2.0\` on npmjs.com will keep showing the OLD README until a new version (0.2.1+) is published. This PR ensures parity from the next release onward — it does not retroactively fix the 0.2.0 page.

If immediate fix is needed for the 0.2.0 page specifically, the only path is to publish \`0.2.1\` (which can be a no-op release containing this README fix). I can prep that PR if needed; otherwise it'll naturally land with the next bug fix or feature release.

## Tier 2 / Tier 3

N/A — markdown + JSON edit. The \`npm pack --dry-run\` above is the operational verification.

## Test plan

- [x] \`npm pack --dry-run\` shows the prepack hook firing and synced README in the tarball.
- [x] Both READMEs are byte-identical after the cp.
- [ ] On the next release (whenever that ships), confirm \`https://npmjs.com/package/movehat\` shows the new README content.

### DIFF
--- a/packages/movehat/README.md
+++ b/packages/movehat/README.md
@@ -1,361 +1,214 @@
-# Movehat
+<div align="center">
+  <img src="./packages/movehat/public/movehat.png" alt="Movehat" width="160"/>
 
-> Hardhat-like development framework for Movement L1 smart contracts
+  # Movehat
 
-[![npm version](https://badge.fury.io/js/movehat.svg)](https://www.npmjs.com/package/movehat)
-[![License: MIT](https://img.shields.io/badge/License-MIT-yellow.svg)](https://opensource.org/licenses/MIT)
+  **A Hardhat-like development framework for Movement L1 smart contracts.**
+
+  Write your tests and deployment scripts in TypeScript while building Move contracts.
+
+  [![NPM Version](https://img.shields.io/npm/v/movehat)](https://www.npmjs.com/package/movehat)
+  [![License](https://img.shields.io/npm/l/movehat)](./LICENSE)
+
+  **[Full documentation](https://movehat.org/)**
+
+</div>
+
+---
 
 ## Features
 
-- **Local Node Testing** - Run a real Movement blockchain locally for testing (just like Hardhat!)
-- **Dual Testing Modes** - Choose between `local-node` (full blockchain) or `fork` (read-only snapshot)
-- **Auto-Deploy in Tests** - Contracts deploy automatically when tests run - zero manual setup
-- **setupTestFixture Helper** - High-level API for test setup with type-safe contracts
-- **Auto-detection of Named Addresses** - Automatically detects and configures addresses from Move code
-- **Quick Start** - Scaffold new Move projects in seconds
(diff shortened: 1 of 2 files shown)


gilbertsahumada (commented):
## §8 Self-review

**🔴 Blockers**: none.

**🟡 Worth fixing**:
- _None_ that block merge. One judgment-level note: \`prepack\` is the right lifecycle hook for both \`npm pack\` and \`npm publish\`, but it's slightly less obvious than a workflow step. Documented inline in the PR body + commit message; future maintainers will see the script in package.json scripts list.

**🟢 Nits**: none.

**Considered + rejected**:
- Adding the cp as an explicit \`.github/workflows/publish.yml\` step: works but duplicates the responsibility. Local \`pnpm test:smoke\` (which calls \`npm pack\`) would still see stale README. \`prepack\` hits both CI and local with one mechanism.
- Symlinking \`packages/movehat/README.md\` → \`../../README.md\`: npm dereferences symlinks during pack; the tarball would contain a copy anyway. Adds a layer of confusion without solving a real problem.

**Pre-merge checklist** (§8):
- [x] §8 review posted.
- [x] No 🔴 / 🟡 findings to resolve.
- [x] Tier 2 / Tier 3 N/A documented.
- [ ] CI green confirms (only third-party under new scope).

Ready to merge once CI completes. Under the new CI scope (#186), only third-party checks run on PRs to develop — full internal CI will fire on the batch develop→main.

PR #194, opened 2026-05-15
fix(dogfood): make Tier B test pass on Apple Silicon + Linux x86_64

gilbertsahumada:
## Summary

Empirically validated PR #193's dogfood test by running it. Six sequential issues surfaced — each documented and fixed in this PR. The dogfood test now passes end-to-end (steps 1–9 green) on Apple Silicon via Docker / Rosetta.

| # | Issue | Fix |
|---|---|---|
| 1 | Rosetta error: \`/lib64/ld-linux-x86-64.so.2\` failed to open under Docker on Apple Silicon. | \`platform: linux/amd64\` in the docker-compose service forces explicit Rosetta emulation. |
| 2 | Movement CLI binary required glibc 2.38+ / CXXABI 1.3.15; Debian Bookworm ships glibc 2.36. | Switched base image from \`node:20-bookworm\` to \`node:20-trixie-slim\` (glibc 2.41 / gcc 14). CI ubuntu-latest 24.04 already satisfies. |
| 3 | \`movement move build\` couldn't fetch \`aptos-core\` git dependency because \`git\` wasn't in the slim image. | Added \`git\` to \`apt-get install\`. |
| 4 | \`aptos-core\` clone exceeded movehat's 2-min runCli timeout under Rosetta. | Pre-warm the \`AptosFramework\` dep cache at Docker build time using a minimal \`scripts/warm/\` package with the same dep. Bakes \`/root/.move/\` into the image; runtime compiles are fast. |
| 5 | \`movehat init /tmp/dogfood-project\` used the full path as the Move package name, producing \`name = "/tmp/dogfood-project"\` in Move.toml — invalid identifier. | Orchestrator now \`cd\`s to \`/tmp\` and runs \`movehat init dogfood-project\` (bare identifier). |
| 6 | \`movement node run-local-testnet\` crashes on Linux x86_64 with the upstream MintFunder \`ENOT_APTOS_FRAMEWORK_ADDRESS\` bug — same one CI gates harness-local behind \`MOVEHAT_SKIP_LOCAL_NODE\`. | Step 8 pivoted from \`movehat test --ts\` (spawns local node) to \`movehat test --move\` (Move unit tests, no node). Move-level \`test_reset()\` injected alongside the entry function so the user-authored \`reset()\` is still validated end-to-end. TypeScript Harness.createLocal coverage continues to live in the integration suite running on macOS hosts (arm64 native, no Rosetta — upstream bug doesn't manifest). |

## Files

[...]

PR #193, opened 2026-05-15
chore: add Tier B end-to-end dogfood test

gilbertsahumada:
## Summary

Adds a Docker-based dogfood test that exercises the full user journey from \`npm install\` to running tests against a real local Movement node, plus fork-mode invariant validation against testnet. Surfaces drift in the install → scaffold → compile → test → fork flow that the existing per-piece tests (\`test:smoke\`, \`test:e2e\`, \`test:integration\`) don't catch when composed end-to-end with a user-authored Move edit.

Tracking issue: [#192](https://github.com/gilbertsahumada/movehat/issues/192) (writable-fork support for future Tier C step).

## Files

| File | Purpose |
|---|---|
| \`Dockerfile.dogfood\` (new) | Node 20 bookworm + real Movement CLI installed via SHA256-pinned tarball (same revision as \`ci.yml\`). Pin must rotate in lockstep with \`MOVEMENT_CLI_SHA256_LINUX_X64\` in \`.github/workflows/ci.yml\`. |
| \`scripts/dogfood-test.sh\` (new, executable) | 9-step orchestrator that runs inside the container. |
| \`scripts/dogfood-fork-test.ts\` (new) | Standalone TS script invoked by the orchestrator at step 9 to assert fork-mode invariants. |
| \`docker-compose.test.yml\` (modified) | New \`test-dogfood\` service bind-mounting \`.\` at \`/workspace\`. |
| \`package.json\` (modified) | New \`docker:test:dogfood\` script. |
| \`TESTING.md\` (modified) | Documents the new flow under "Docker Testing". |

## The 9-step orchestrator

1. **Verify environment** — node / movement / pnpm versions.
2. **Build + pack movehat** from \`/workspace/packages/movehat\` via \`npm pack\`.
3. **Install globally** — \`npm install -g\` on the freshly-packed tarball.
4. **Scaffold** — \`movehat init /tmp/dogfood-project\`.
5. **Extend the template Counter with a user-authored \`reset()\`** entry function and a matching mocha test. Validates the framework supports real contributor edits, not just the canned template.
6. **Install project deps** — \`npm install\` in the scaffolded project.
7. **Compile** — \`movehat compile\` against the real Movement CLI.
8. **Run mocha** — \`movehat test --ts\` spawns a real local Movement node via \`Harness.createLocal\`, autoDeploys, runs tests including the user-authored \`reset()\` test. Asserts >= 1 test passes.
[...]

PR #196, opened 2026-05-15
chore(gitignore): ignore runtime artifacts that leak into repo

gilbertsahumada:
## Summary

Three sources of untracked-file noise were filling up the working tree (10k+ files surfaced in IDE file listings):

| Path | Source | Size |
|---|---|---|
| `.pnpm-store/` | pnpm CAS cache. Leaks into repo root when `pnpm install` runs inside the dogfood Docker container — no access to host `~/.local/share/pnpm/store/`, so pnpm falls back to CWD which under bind-mount is the host repo. | 723MB / 27k files |
| `Move.lock` | Generated by `movement move build` (specifically in `scripts/warm/` from the dogfood image pre-warm step, but applies to any Move package). | ~2KB |
| `.claude/` | Claude Code session state — per-machine runtime, not project content. | <1KB |

Also removed `packages/movehat/public/movehat_move.png` — uncommitted since May 6, not referenced in any `.md`/`.mdx`/`.tsx`. Orphan.

## Why now

Surfaced while preparing the develop→main batch for Tier B dogfood (PRs #193 + #194). The pnpm-store leak is a direct side-effect of the new Docker dogfood test, so the ignore rule belongs in the same milestone.

## Test plan

- [x] `git status` clean after applying.
- [x] Existing tracked files unaffected (only `.gitignore` modified).

## Tier 2 / Tier 3

N/A — `.gitignore` change + orphan asset removal. No source / runtime / packaging behavior.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -21,6 +21,15 @@ packages/docs/content/docs/api/reference/
 /tmp/
 .movehat/
 
+# Move build artifacts (leak from Docker bind-mount when movement CLI runs inside container)
+Move.lock
+
+# pnpm CAS cache (leaks into repo root when pnpm runs inside Docker without access to ~/.local/share/pnpm/store)
+.pnpm-store/
+
+# Claude Code session state
+.claude/
+
 # Local reference 
 /catapulta/
 


gilbertsahumada (commented):
## §8 Self-review

Pure `.gitignore` change + orphan asset removal. Re-read confirms:

- Three ignore rules each map 1:1 to a documented leak source (PR body table).
- `.claude/` ignore won't hide project-level Claude Code config the user might want to commit later; if/when needed we can use `!.claude/<file>` to whitelist.
- `Move.lock` is project-agnostic (no path prefix) — correct, since it can appear in any nested Move package (the warm one, future user packages dropped in `examples/`, etc.).
- `.pnpm-store/` only covers root-level; pnpm normally doesn't create nested stores so a global `**/.pnpm-store/` would be over-broad.

No 🔴 / 🟡 / 🟢 findings. Ready to merge once Socket finishes.

PR #198, opened 2026-05-15
feat(init): sanitize project name as Move identifier

### LINKED ISSUE #195: movehat init: validate project name as a Move identifier (currently accepts paths and produces invalid Move.toml)
## Summary

\`movehat init <name>\` uses the \`<name>\` argument as both the directory path to create AND the Move package name written to \`move/Move.toml\`. The command accepts any string without validating that it's a valid Move identifier. When the argument is a path (e.g. \`/tmp/my-project\`) or contains characters illegal in a Move identifier (slashes, dots, leading digits, etc.), the resulting \`Move.toml\` is malformed and surfaces later as a cryptic compile error far removed from the actual cause.

## Reproduction

```bash
# Pass a path instead of a bare name
movehat init /tmp/my-project
cd /tmp/my-project
cat move/Move.toml
```

Observed \`move/Move.toml\`:

```toml
[package]
name = \"/tmp/my-project\"      # invalid: contains slashes
version = \"1.0.0\"
authors = []
...
```

Now try to compile:

```bash
movehat compile
```

Observed output (cleaned):

```
ℹ Compiling Move contracts...
  Move directory: /tmp/my-project/move
  Detected addresses: counter
  Auto-assigned (0xcafe): counter

{
  \"Error\": \"Move compilation failed: No such file or directory (os error 2)\"
}
Compiling, may take a little while to download git dependencies...
UPDATING GIT DEPENDENCY https://github.com/movementlabsxyz/aptos-core.git
INCLUDING DEPENDENCY AptosFramework
INCLUDING DEPENDENCY AptosStdlib
INCLUDING DEPENDENCY MoveStdlib
BUILDING /tmp/my-project

✘ Compilation failed: movement move build exited with code 1
```

The actual problem — \"package name is not a valid identifier\" — is nowhere in the error message. The diagnostic surfaces as a generic OS-level \"No such file or directory\" because Movement CLI's internal parser fails partway through and tries to read a derived file path that includes the malformed name.

## Expected behavior

\`movehat init\` should either:

1. **Validate the argument** as a valid Move identifier and refuse with a clear error message:
   ```
   ✘ Invalid project name: '/tmp/my-project'
   Move package names must match: [a-zA-Z_][a-zA-Z0-9_]*
   ```

2. **Derive a safe package name from the path basename**, accepting the argument as a path but normalizing the package name:
   - Take \`basename(arg)\` (e.g. \`my-project\` from \`/tmp/my-project\`)
   - Strip / sanitize illegal characters (hyphens → underscores, etc.)
   - Use the sanitized form for \`Move.toml\`'s \`name\` field

3. **Both**: accept the argument as a path/name, derive a sane default, and warn the user if the original differed from the sanitized form:
   ```
   ⚠ Sanitizing package name: 'my-project' → 'my_project' (Move identifiers cannot contain '-')
   ```

Option 3 is the most user-friendly and matches what \`cargo new\` does for Rust project names.

## Move identifier rules (reference)

From the Move language specification: identifiers must match \`[a-zA-Z_][a-zA-Z0-9_]*\`. Specifically:
- Must start with a letter or underscore
- May contain letters, digits, and underscores
- May NOT contain hyphens, slashes, dots, or other punctuation

The TOML spec allows arbitrary strings in quoted keys/values, so Move.toml technically lets you write \`name = \"/tmp/my-project\"\` — but Movement CLI's downstream consumers (the compiler, dependency resolver, etc.) all assume the name is a valid identifier and break in subtle ways when it isn't.

## Suggested implementation

The init command lives at \`packages/movehat/src/commands/init.ts\`. The flow is roughly:

1. Resolve the target directory (current code: uses \`<name>\` argument directly)
2. Copy the template tree
3. Substitute \`{{projectName}}\` in template files (\`Move.toml\`, \`package.json\`, etc.)

The fix touches step 1 and step 3:

```typescript
function sanitizePackageName(input: string): { dirName: string; packageName: string; sanitized: boolean } {
  // Accept paths: split out the basename
  const basename = path.basename(path.resolve(input));

  // Sanitize for Move identifier rules
  const sanitized = basename.replace(/[^a-zA-Z0-9_]/g, '_');
  const final = /^[a-zA-Z_]/.test(sanitized) ? sanitized : \`pkg_\${sanitized}\`;

  return {
    dirName: input,                     // preserve the user's intent for the directory
    packageName: final,                  // a guaranteed-valid Move identifier
    sanitized: final !== basename,       // did we have to change anything?
  };
}
```

Then surface a warning when \`sanitized\` is true so the user knows what happened, and use \`packageName\` for the \`{{projectName}}\` substitution in templates.

## Where this was surfaced

While building the Tier B dogfood test (PR #193, fixes in PR #194). The orchestrator script ran:

```bash
movehat init /tmp/dogfood-project
```

…which broke compile with the \"No such file or directory\" error above. Took ~3 iterations of debug-dumping intermediate files to figure out the package name in \`Move.toml\` was the actual problem. A new user hitting this would have a much harder time diagnosing — the error message gives no hint that the name is the culprit.

Tracking workaround in the dogfood orchestrator (\`scripts/dogfood-test.sh\`): \`cd\` to the parent dir and invoke \`movehat init <basename>\`. That sidesteps the bug but is exactly the kind of thing \`movehat init\` itself should handle.

## Suggested scope

- Pure UX / DX improvement. No public API change.
- One-file change in \`packages/movehat/src/commands/init.ts\`.
- Add a unit test under \`packages/movehat/src/commands/__tests__/init.test.ts\` covering:
  - Bare name → unchanged.
  - Path argument → basename used as package name.
  - Hyphens → underscores with warning.
  - Leading digit → \`pkg_\` prefix with warning.
- Worth a small entry in \`CHANGELOG.md\` Unreleased → Changed.

## Priority

Low / not urgent. The workaround is trivial (\"don't pass a path\") and the failure is loud (compile fails immediately). But the error message is unhelpful enough that fixing this would save a real-user-onboarding paper cut.

gilbertsahumada:
## Summary

Closes #195.

\`movehat init <name>\` accepted any string as both the directory path and the Move package name, with no validation. Passing a path (e.g. \`/tmp/my-project\`) or a hyphenated identifier wrote that string verbatim into \`move/Move.toml\`'s \`name\` field, producing an invalid Move identifier and a cryptic compile failure downstream.

Sanitize-and-warn (cargo-style): filesystem directory = user input as-is; the Move package name is auto-derived from the basename and sanitized into a valid identifier; \`package.json\` keeps the original (npm allows hyphens).

## Behavior matrix

| Input | Directory | \`package.json\` name | \`Move.toml\` name | Warning |
|---|---|---|---|---|
| \`my_project\` | \`my_project\` | \`my_project\` | \`my_project\` | — |
| \`my-project\` | \`my-project\` | \`my-project\` | \`my_project\` | yes |
| \`/tmp/my-project\` | \`/tmp/my-project\` | \`my-project\` | \`my_project\` | yes |
| \`123abc\` | \`123abc\` | \`123abc\` | \`pkg_123abc\` | yes |
| \`.\` / \`..\` / \`\` / \`/\` | — | — | — | error (exit 1) |

## Changes

- \`packages/movehat/src/commands/init.ts\` — \`resolveProjectNames\` pure helper + \`InvalidProjectNameError\`. Wired into \`initCommand\`. Dual replacement map passed to \`copyDir(move, ...)\` so the same template position receives different values for \`package.json\` (npm-friendly) vs \`Move.toml\` (strict identifier).
- \`packages/movehat/src/templates/move/Move.toml\` — placeholder renamed \`{{projectName}}\` → \`{{movePackageName}}\`.
- \`packages/movehat/src/commands/__tests__/init.test.ts\` — table-driven unit tests for the helper (8 happy cases + 6 rejection cases) plus integration tests for hyphen, path, leading-digit, and dot/empty inputs.
- \`scripts/dogfood-test.sh\` — removes the \`cd parent; init basename\` workaround introduced in PR #194. \`init\` now handles paths natively.
- \`CHANGELOG.md\` — Unreleased → Changed entry.

## Verification (Tier 1)

| Check | Result |
|---|---|
| \`pnpm test\` (full unit suite) | 417/417 green (+21 new tests) |
| \`src/commands/init.ts\` coverage | **97.59% lines / 93.1% branches** (gate ≥80% per \`vitest.config.ts:49\`) |
| \`pnpm check:example\` | green (movehat builds; example tsc --noEmit clean) |

## Tier 2 / Tier 3

- Tier 2: \`pnpm test:smoke\` and \`pnpm test:example\` — not run in this PR (no packaging change; \`bin\` shim untouched, no template file additions/removals, only a placeholder rename inside a template). Will be exercised on the next packaging-touching PR or pre-publish.
- Tier 3: N/A — no \`bin/**\` change.

## Risks

- Existing user calling \`movehat init my-project\` non-interactively in CI would see a new warning line where there was none. Warning, not error. CI keeps working. The behavior change in \`Move.toml\` is the bug fix.
- Placeholder rename \`{{projectName}}\` → \`{{movePackageName}}\` in \`Move.toml\` is contained to the template (consumed only by \`init.ts\`); not a public surface.

## Out of scope (deferred)

- npm name lowercasing / reserved-word check (Movement CLI catches at compile, npm at publish).
- \`--package-name\` flag to override auto-derivation. Auto-derivation handles >95% of cases; flag would be useful only for collisions between basename and an existing Move identifier, which is rare and easy to work around by renaming the directory.

## Test plan

- [x] \`pnpm test\` 417/417 green.
- [x] init.ts per-file coverage 97.59% (≥80% gate).
- [x] \`pnpm check:example\` green.
- [x] Dogfood orchestrator workaround removed; CI on this PR will validate the rest.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,17 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Changed
 
+- `movehat init <name>` now sanitizes the project name into a valid Move
+  identifier when writing `move/Move.toml`, preserving the original name for
+  `package.json` and the directory. Invalid characters (hyphens, slashes,
+  dots) are replaced with underscores; names starting with a digit are
+  prefixed with `pkg_`. A warning is printed when sanitization changes the
+  name. Names that resolve to nothing usable (`.`, `..`, empty, only
+  separators) are now rejected with a clear error. Previously, passing a
+  path like `/tmp/my-project` produced a malformed `Move.toml` and a cryptic
+  `"No such file or directory"` compile failure. Closes
+  [#195](https://github.com/gilbertsahumada/movehat/issues/195).
+
 ### Deprecated
 
 ### Removed
--- a/packages/movehat/src/commands/__tests__/init.test.ts
+++ b/packages/movehat/src/commands/__tests__/init.test.ts
@@ -13,7 +13,8 @@ vi.mock("../../helpers/banner.js", () => ({
   printMovehatBanner: () => undefined,
 }));
 
-const { default: initCommand } = await import("../init.js");
+const initModule = await import("../init.js");
+const { default: initCommand, resolveProjectNames, InvalidProjectNameError } = initModule;
 
 /**
  * Strategy: real tmpdir + real fs ops (matches §6.1's example-as-canonical
@@ -28,10 +29,38 @@ const { default: initCommand } = await import("../init.js");
  * to `src/templates/` cleanly. No bundling step required.
  */
 
+describe("resolveProjectNames", () => {
+  const cases: Array<[
+    string,
+    { dirName: string; npmName: string; moveName: string; sanitized: boolean }
+  ]> = [
+    ["my_project", { dirName: "my_project", npmName: "my_project", moveName: "my_project", sanitized: false }],
+    ["my-project", { dirName: "my-project", npmName: "my-project", moveName: "my_project", sanitized: true }],
+    ["/tmp/my-project", { dirName: "/tmp/my-project", npmName: "my-project", moveName: "my_project", sanitized: true }],
+    ["123abc", { dirName: "123abc", npmName: "123abc", moveName: "pkg_123abc", sanitized: true }],
+    ["_underscore", { dirName: "_underscore", npmName: "_underscore", moveName: "_underscore", sanitized: false }],
+    ["UPPER", { dirName: "UPPER", npmName: "UPPER", moveName: "UPPER", sanitized: false }],
+    ["  spaced  ", { dirName: "spaced", npmName: "spaced", moveName: "spaced", sanitized: false }],
+    ["with spaces", { dirName: "with spaces", npmName: "with spaces", moveName: "with_spaces", sanitized: true }],
+  ];
+
+  it.each(cases)("derives names for %j", (input, expected) => {
+    expect(resolveProjectNames(input)).toEqual(expected);
+  });
+
+  it.each(["", "   ", ".", "..", "/", "////"])(
+    "rejects %j as an invalid project name",
+    (input) => {
+      expect(() => resolveProjectNames(input)).toThrow(InvalidProjectNameError);
+    }
+  );
+});
+
 describe("initCommand", () => {
   let tmpParent: string;
   let origCwd: string;
   let exitSpy: ReturnType<typeof vi.spyOn>;
+  let warnSpy: ReturnType<typeof vi.spyOn>;
 
   beforeEach(() => {
     promptsMock.mockReset();
@@ -48,6 +77,7 @@ describe("initCommand", () => {
       }) as never);
     vi.spyOn(console, "log").mockImplementation(() => undefined);
     vi.spyOn(console, "error").mockImplementation(() => undefined);
+    warnSpy = vi.spyOn(console, "warn").mockImplementation(() => undefined);
   });
 
   afterEach(() => {
@@ -59,9 +89,9 @@ describe("initCommand", () => {
   });
 
   it("happy path: scaffolds a project from the canonical template", async () => {
-    await initCommand("my-test-project");
+    await initCommand("my_test_project");
 
-    const target = join(tmpParent, "my-test-project");
+    const target = join(tmpParent, "my_test_project");
     expect(existsSync(target)).toBe(true);
     // Canonical files at the project root.
     expect(existsSync(join(target, "package.json"))).toBe(true);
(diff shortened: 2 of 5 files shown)


gilbertsahumada (commented):
## §8 Self-review

Re-read the diff line-by-line. Findings:

### 🔴 Blockers

None.

### 🟡 Worth noting

- **Dogfood workaround bundled vs deferred**: per the plan's open question, the `scripts/dogfood-test.sh` workaround removal was bundled into this PR rather than deferred. Rationale: keeping the fix and its consumer aligned in the same commit makes bisect cleaner — anyone scrolling back to "when did the dogfood test stop using the workaround" lands on the same commit that made the workaround unnecessary. The alternative (separate follow-up PR) would have left the orchestrator carrying a comment that contradicts the just-shipped behavior. Bundled wins.

- **Test happy-path renamed from `my-test-project` to `my_test_project`**: the pre-existing happy-path test was passing a hyphenated name and only checking `package.json` (so the malformed `Move.toml` never surfaced). I renamed it to a valid identifier to keep the canonical happy-path silent (no sanitization warning), and added a dedicated `"sanitizes Move.toml for hyphenated names but keeps package.json original"` test for the hyphen case. Net: the test surface is broader, and the "happy" path is now genuinely silent.

### 🟢 Nits / deliberately deferred

- **npm name not lowercased**: `npmName = path.basename(...)` preserves case. npm's modern policy is "all lowercase" but it only warns at publish time, not at install/local-use. Adding lowercasing here would silently rename users' input, which is more surprising than warning when it matters (at `npm publish`). Out of scope.

- **No reserved-word check** (`signer`, `address`, `module` etc.): Movement CLI catches these at compile time with a clearer error than the path/hyphen case (which gave `"No such file or directory"`). Adding a parallel allowlist in `init.ts` would duplicate Move compiler logic and drift. Out of scope.

- **No `--package-name` flag**: handled in plan's "Out of scope". Auto-derivation covers >95% of cases; the edge case of "basename collides with an existing Move identifier in the same workspace" is rare and trivially fixable by renaming the directory.

### Pre-merge checklist (CLAUDE.md §8)

- [x] §8 self-review posted (this comment).
- [x] No 🔴 / 🟡 findings require changes on this branch; the 🟡 is the conscious bundling decision, not a deferred fix.
- [x] Tier 2 / Tier 3 marked N/A in PR body (no packaging change; `bin` shim, template file count, dist contents all unchanged — only a placeholder rename inside an existing template file).
- [x] 4/4 CI checks green (CodeRabbit, GitGuardian, Socket x2). Internal CI doesn't run on PRs to `develop` under the post-PR-#186 scope.
- [x] `mergeable: MERGEABLE`.

Ready to merge.

Issue #243, opened 2026-05-19
ForkServer: proxy /v1/view to upstream RPC + clarify fork-mode write rejection


gilbertsahumada:
## Problem

`packages/movehat/src/fork/server.ts` only routes 4 GET endpoints (`/v1/`, `/v1/accounts/:addr`, `/v1/accounts/:addr/resource/:type`, `/v1/accounts/:addr/resources`). Any other path falls through to `send404(res, 'Endpoint not found: ...', 'endpoint_not_found')` at line 220.

But when `Harness.createFork(network)` runs (via `setupLocalTesting` → `setupWithFork` in `packages/movehat/src/helpers/setupLocalTesting.ts:355-368`), the runtime's RPC `url` is wired to `http://localhost:${forkPort}/v1` — i.e. **the ForkServer is the runtime's only RPC.** Consequences:

- `harness.runViewFunction(...)` POSTs to `/v1/view` → ForkServer returns 404
- `MoveContract.view(...)` does the same
- `MoveContract.call(...)` POSTs to `/v1/transactions` → ForkServer returns 404 (and there's no harness-level guard on `.call()` either; the existing fork-mode guards in `Harness.ts` only cover `deployCodeObject` / `upgradeCodeObject` / `runMoveScript`)

## What the existing docs claim

`packages/docs/content/docs/guides/networks-and-modes.mdx` table (line 14) claims:

> Read real testnet / mainnet state (writes simulate locally) | `createFork(network)` | reads live, writes rejected

And the per-mode section lists `runViewFunction` and `MoveContract.view` as "Allowed". Both claims are wrong against the current ForkServer surface.

## Two changes proposed

**1. Add `/v1/view` proxy to ForkServer** — pass-through to upstream RPC (cached if useful, but read-only views are idempotent so caching isn't critical). Unblocks `runViewFunction` in fork mode.

**2. Either guard `MoveContract.call` at the harness Proxy in fork mode, OR add `/v1/transactions` simulate-only handling.** The latter is the more useful direction — it's what EVM tools (Forge `--fork-url`, Hardhat fork) do — but it's also a lot more code. The guard approach is a 5-line change that fails loudly with a clear message instead of an opaque 404.

Mitigation in the meantime: docs corrected in the M8.1 PR to accurately describe what works today.

## DoD

- [ ] `/v1/view` POST handled in `packages/movehat/src/fork/server.ts`
[...]

Issue #238, opened 2026-05-19
[Roadmap] M8: KPI 2 delivery — TypeDoc polish + tutorials + maintenance policy


gilbertsahumada:
## Goal

Close the formal KPI 2 commitments listed in `grant/KPI1.md` §E.2: tutorial expansion, TypeDoc structural cleanup, and maintenance cadence formalization. Produces `grant/KPI2.{md,pdf}` plus a Telegram message to Move Industries.

## Sub-issues

- [ ] #239 — M8.1 Tutorial: fork-mode testing
- [ ] #240 — M8.2 Tutorial: deploying to live networks
- [ ] #241 — M8.3 Tutorial: CI integration + reference workflow
- [ ] #242 — M8.4 TypeDoc structural cleanup + MAINTENANCE.md + KPI2 report

## Definition of Done

- [ ] `packages/docs/content/docs/guides/tutorial-fork-testing.mdx` live on movehat.org
- [ ] `packages/docs/content/docs/guides/tutorial-deploy-live.mdx` live on movehat.org
- [ ] `packages/docs/content/docs/guides/tutorial-ci.mdx` live on movehat.org
- [ ] `examples/counter-example/.github/workflows/ci.yml` committed and identical to the tutorial snippet
- [ ] `movehat.org/docs/api/reference/` sidebar groups symbols by category (Harness / Account / Contract / Fork / Deployment Helpers / Errors), not flat alphabetical
- [ ] `MAINTENANCE.md` at repo root with release cadence + triage SLA + deprecation policy
- [ ] `contributing/maintenance.mdx` exists and links from `contributing/index.mdx` and `README.md`
- [ ] `grant/KPI2.pdf` generated, sign-off date set
- [ ] Telegram draft in `grant/email-draft-kpi2.md` references `KPI2.pdf` + delivery commit hash
- [ ] CI green on `develop → main` batch PR
- [ ] `docs-deploy.yml` succeeds on batch merge; all three new tutorial pages return HTTP 200

## Scope discipline

Out of scope (file separately): issue #235 mocha root-hooks, issue #223 console.* sweep, factory rename, `createLive`-in-test footgun-guard, anvil-lite exploration, mainnet deploy as evidence.

## Effort

~5 working days + 2 days for batch / report / send.

## Workflow

[...]

Issue #192, opened 2026-05-15
Fork system: support transaction submission against fork state


gilbertsahumada:
## Context

The current fork system (`Harness.createFork`, `ForkManager`, `ForkServer`) is **read-only by design**: it caches resources from a remote network as JSON snapshots and serves them back to the SDK. Transaction submission, account creation, and state mutation are intentionally rejected — \`harness.deployCodeObject\` / \`upgradeCodeObject\` / \`runMoveScript\` throw a clear error when called on a fork-mode harness.

This works well for read-flow testing (\"what does this view function return against current testnet state?\") and write-rejection testing (\"does my code correctly bail when given a read-only handle?\"), but it does **not** support a common Foundry-style use case:

> *\"Fork mainnet, then deploy a new contract into that forked state and test how it interacts with existing on-chain resources.\"*

In that flow, the fork is the substrate for a local Move VM that processes the test transactions but starts from real network state.

## What 'writable fork' would unlock

- Test contract upgrades against real on-chain state (e.g., \"if I upgrade module X with this new logic, does it still interoperate with the existing testnet token Y?\")
- Reproduce production bugs locally by forking the block where they happened and replaying transactions
- Integration tests that need real testnet resources but should NOT touch the testnet itself
- Faster test cycles than full \`createLocal\` (no genesis setup; the fork starts at a real block height)

This is the semantic Foundry's \`anvil --fork-url ...\` and Hardhat's \`hardhat_reset\` + \`forking.url\` provide. Movement currently has no equivalent that's wrapped at a framework level.

## Implementation directions (rough)

Multiple paths possible; tradeoffs in each:

1. **Local Movement node initialized from fork snapshot**: spawn a real \`movement node run-local-testnet\` but seed the local Move state from the fork's cached resources. Most faithful to production, but requires Movement-CLI-level genesis customization which may not be supported upstream.

2. **Extend \`ForkServer\` with a transaction-submission endpoint**: handle \`/transactions\` POSTs by deserializing the BCS payload, executing it against a local Move VM hydrated from the cached state, and persisting the resulting state mutations back into the fork's JSON storage. Self-contained but means re-implementing transaction execution semantics inside Movehat.

3. **Hybrid**: run a local Movement node with empty genesis, then pre-populate it via the fork's snapshot data on startup (via \`setResource\` over the local node's REST API). Requires the local Movement node to expose a state-mutation endpoint.

[...]

PR #244, opened 2026-05-19
docs(m8.1): fork-mode testing tutorial + networks-and-modes accuracy fix

### LINKED ISSUE #239: [M8.1] Tutorial: fork-mode testing
Tracks #238

## Goal

Ship a step-by-step tutorial that walks the user through using `Harness.createFork(network)` to read real mainnet state in tests without spending gas.

## Deliverable

`packages/docs/content/docs/guides/tutorial-fork-testing.mdx` (~150 LoC) following the `guides/multi-contract.mdx` numbered-steps + troubleshooting + cross-refs format.

## Content outline

1. Prerequisites (quickstart done, `MOVEHAT_NETWORK` understood)
2. What we're building — a test reading a known mainnet contract's state
3. Step 1 — Configure `createFork("mainnet")`; what gets cached vs proxied
4. Step 2 — Write the read-only assertion via `runViewFunction` against a real mainnet view (use `0x1::aptos_coin` framework views — stable, not third-party)
5. Step 3 — Demonstrate the write-rejection guard (try `.call`, observe `HarnessForkWriteError`)
6. Step 4 — Run + inspect: `pnpm test:ts`, fork-server lifecycle in `-v` mode
7. Troubleshooting — stale fork cache, RPC rate limits, mainnet endpoint flapping
8. Cross-refs to `guides/fork`, `guides/networks-and-modes`

## Nav

- Add `tutorial-fork-testing` to `packages/docs/content/docs/guides/meta.json` after `multi-contract`
- Add `guides/tutorial-fork-testing` to `packages/docs/content/docs/meta.json` in the Guides block

## DoD

- [ ] New MDX file exists and renders cleanly (`cd packages/docs && npx next build` passes)
- [ ] Both `meta.json` files updated
- [ ] Tier 1 typecheck still passes (`pnpm check:example`)
- [ ] §8 self-review posted before merge
- [ ] M8 ROADMAP DoD bullet ticked in the same PR (§7)
- [ ] PR merged to `develop`

## Out of scope

The page only documents the existing `createFork` API. No new features. Issue #192 (simulated-writes-against-fork-state) is referenced as a future-work pointer, not implemented here.

gilbertsahumada:
## Summary

- Adds `packages/docs/content/docs/guides/tutorial-fork-testing.mdx` — a step-by-step tutorial that walks the reader through using `Harness.createFork("mainnet")` to read real mainnet account + resource state in a TypeScript test, demonstrates the synchronous deploy-rejection guard, and explains the snapshot-cache behaviour.
- Corrects two false claims in `packages/docs/content/docs/guides/networks-and-modes.mdx` shipped with the KPI1 delivery: the previous version listed `runViewFunction`, `MoveContract.view`, and `getContract(...).view` as "Allowed" in fork mode. They are not. `ForkServer` (`packages/movehat/src/fork/server.ts:220`) routes 4 GET endpoints and returns 404 for everything else, including `/v1/view` and `/v1/transactions`.
- Files a follow-up issue #243 for the ForkServer `/v1/view` proxy gap.
- Adds `tutorial-fork-testing` to both `meta.json` files; ticks the M8.1 DoD row in `ROADMAP.md`.

## What I verified

- `cd packages/docs && npx next build` → clean; 71 routes (was 70).
- `pnpm typecheck:example` → green.
- Pre-push hook (`pnpm check:example`) → green.
- The tutorial's code paths align with the fork server's actual endpoint surface (account info, single resource, resources bundle). The deploy-rejection step uses the harness Proxy's existing guard, which is genuinely typed.

## What I did **not** verify (honest gaps)

- End-to-end execution of the tutorial against a live mainnet fork. The tutorial is a docs deliverable, not part of `examples/counter-example/`, so it is not exercised by `pnpm test:example`. The patterns are sourced from the implementation, not from a live run.
- The "ForkServer 404 on /v1/view" claim is from source reading; if it turns out the runtime falls back to upstream RPC when the fork server 404s, the tutorial would actually under-promise. The advisor recommended a runtime check; I read the test file and source instead — happy to do the runtime check in a follow-up if reviewer prefers.

## Out of scope

- Implementing the `/v1/view` proxy in `ForkServer` (tracked in #243).
- Implementing the `/v1/transactions` simulate path (tracked separately under the writable-fork umbrella #192).
- The other three M8 tutorials (M8.2 deploy-live, M8.3 CI, M8.4 TypeDoc + maintenance).

## Test plan

- [x] `pnpm typecheck:example`
- [x] `cd packages/docs && npx next build`
- [x] Pre-push `pnpm check:example`
- [ ] After merge to `develop` → `main` batch: `curl -s -o /dev/null -w "%{http_code}\n" https://movehat.org/docs/guides/tutorial-fork-testing` returns 200

Closes #239
Tracks #238

### DIFF
--- a/ROADMAP.md
+++ b/ROADMAP.md
@@ -283,6 +283,33 @@ Follow-up issues filed during M4 (all upstream Movement CLI / SDK, deferred to M
 - [ ] Distributed opportunistically across milestones (e.g., #41 fits naturally in M1, #44 in M0)
 - [ ] No critical (`security` + `bug`) audit issues left open at the time of M6
 
+### M8 — KPI 2 delivery: TypeDoc polish + tutorials + maintenance policy (~5 days, issue [#238](https://github.com/gilbertsahumada/movehat/issues/238))
+**Goal**: Close the formal KPI 2 commitments listed in `grant/KPI1.md` §E.2.
+
+**Definition of Done**:
+- [x] `packages/docs/content/docs/guides/tutorial-fork-testing.mdx` live on movehat.org (M8.1)
+- [ ] `packages/docs/content/docs/guides/tutorial-deploy-live.mdx` live on movehat.org
+- [ ] `packages/docs/content/docs/guides/tutorial-ci.mdx` live on movehat.org
+- [ ] `examples/counter-example/.github/workflows/ci.yml` committed and identical to the snippet in the CI tutorial
+- [ ] `movehat.org/docs/api/reference/` sidebar groups symbols by category (Harness / Account / Contract / Fork / Deployment Helpers / Errors), not flat alphabetical
+- [ ] `MAINTENANCE.md` at repo root with release cadence + triage SLA + deprecation policy
+- [ ] `packages/docs/content/docs/contributing/maintenance.mdx` exists and links from `contributing/index.mdx`
+- [ ] `grant/KPI2.pdf` generated, sign-off date set
+- [ ] Telegram draft in `grant/email-draft-kpi2.md` references `KPI2.pdf` + delivery commit hash
+- [ ] CI green on `develop → main` batch PR
+- [ ] `docs-deploy.yml` succeeds on the batch merge; all three new tutorial pages return HTTP 200
+
+**Sub-PRs**:
+
+| Sub | Description | Issue | Status |
+|---|---|---|---|
+| M8.1 | Fork-mode tutorial + `networks-and-modes` accuracy fix | [#239](https://github.com/gilbertsahumada/movehat/issues/239) | in PR |
+| M8.2 | Deploy-to-live tutorial | [#240](https://github.com/gilbertsahumada/movehat/issues/240) | |
+| M8.3 | CI integration tutorial + reference workflow | [#241](https://github.com/gilbertsahumada/movehat/issues/241) | |
+| M8.4 | TypeDoc structural cleanup + MAINTENANCE.md + KPI2 report | [#242](https://github.com/gilbertsahumada/movehat/issues/242) | |
+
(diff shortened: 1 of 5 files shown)


gilbertsahumada (commented):
## Self-review (§8)

### 🔴 Blockers

None.

### 🟡 Worth fixing / explicitly deferring

1. **The "MoveContract.call fails with opaque 404" claim is not runtime-verified.** The advisor flagged this and recommended a 2-minute empirical check; I sourced it from the ForkServer endpoint surface in `packages/movehat/src/fork/server.ts:220` plus the runtime URL wiring in `packages/movehat/src/helpers/setupLocalTesting.ts:362` instead. If a reviewer wants the live check, I can run `Harness.createFork("mainnet")` + a `MoveContract.call(...)` against an arbitrary mainnet address in a follow-up and confirm the actual error shape — happy to either way. Deferring: the source path is unambiguous (the runtime points at the fork server, which has no `/v1/transactions` handler), so the directional claim is correct even if the exact error string differs.

2. **`networks-and-modes.mdx` was shipped wrong with KPI1 and is being corrected here.** The PR description calls this out, but the KPI1 PDF (`grant/KPI1.pdf` already sent to Move Industries) does not mention this guide specifically — only the docs URL. No correction needed to the PDF; the live docs are the source of truth and are being fixed now.

### 🟢 Nits

- The fork-name collision footgun (`test-local` reused across networks) is documented in troubleshooting but the Step 1 example doesn't preempt it (`Harness.createFork` doesn't expose `forkName`). Could surface a "drop down to `setupLocalTesting` if you need parallel forks per network" tip earlier, but that's a separate UX concern.
- Step 5's failing-write example uses `harness.deployCodeObject` because the harness Proxy guard for that is typed and clean. A reader might expect `MoveContract.call` because that's what a normal test would attempt. I noted the call-vs-deploy distinction inline, so the gap is documented.

### Resolution

No 🟡 fixes applied in this PR. Both are explicitly deferred above — the first is verifiable in a follow-up, the second is informational.

PR #245, opened 2026-05-20
feat(m8.0): proxy POST /v1/view from ForkServer to upstream RPC

### LINKED ISSUE #243: ForkServer: proxy /v1/view to upstream RPC + clarify fork-mode write rejection

gilbertsahumada:
## Summary

- Adds `POST /v1/view` passthrough proxy to `ForkServer`. Closes the gap surfaced empirically in PR #244 (`/tmp/fork-gap-demo/demo.mjs`).
- `MovementApiClient` gains a private `post<T>` helper mirroring `get<T>` (TLS/timeout/maxBytes/error wrapping) and a public `view(payload)` method.
- `ForkManager` gains a `forwardView(payload)` passthrough (no caching — view results depend on ledger version + args, and version-aware invalidation does not exist today).
- `ForkServer` gains `handleViewProxy(req, res)` — buffers the body with a 1 MiB ceiling, parses JSON, calls `forwardView`, returns the upstream array verbatim. Errors map cleanly: 413 (body too large), 400 (invalid JSON), 502 (upstream failure).
- Three vitest cases in `src/fork/__tests__/server.test.ts` cover happy path, bad JSON, and upstream failure — using the existing `vi.mock("../api.js")` pattern for offline determinism.
- Docs flipped: `networks-and-modes.mdx` and `tutorial-fork-testing.mdx` now promote view-fn support to "Supported" with a real working example (`0x1::coin::supply<AptosCoin>` against Movement testnet) and remove the stale "Endpoint not found: /v1/view" troubleshooting entry.

Out of scope (stays in #192): `POST /v1/transactions` — writable-fork is materially harder and needs either a local Move VM or upstream `simulate_transaction` integration.

## What I verified

- **Unit tests**: `pnpm --filter movehat test` → 471 passed (5 in the touched `server.test.ts`, including 3 new cases).
- **Coverage gate**: `pnpm --filter movehat test:coverage` → green (per-file gates on critical modules unchanged). `manager.ts` at 96.33%.
- **Tier 1 typecheck**: `pnpm check:example` → green.
- **Docs build**: `cd packages/docs && npx next build` → 71 routes, clean.
- **Empirical** (the important one): re-ran `/tmp/fork-gap-demo/demo.mjs` after rebuilding movehat. `POST /v1/view` now returns **HTTP 200** with real testnet data:

  ```
  HTTP 200  POST /v1/view  (view fn — claimed)
              body: [
                {
                  "vec": [
                    "418947910172177144802"
                  ]
                }
              ]
  ```

  That's the live MOVE total supply in octas from Movement testnet. `POST /v1/transactions` correctly stays at 404 (still #192's territory).

## What I did **not** verify

- Behavior under upstream RPC rate limiting. The proxy forwards the upstream's response verbatim, so a 429 from upstream will surface as a 502 with the original message inside — acceptable.
- BCS-encoded view responses. The endpoint defaults to JSON per the Movement REST spec; if a caller sets `Accept: application/x-bcs`, our header-passthrough is incomplete (we set `Content-Type: application/json` on the outgoing request and don't forward `Accept`). Filing as a follow-up if any caller actually needs it.

## Test plan

- [x] `pnpm --filter movehat test`
- [x] `pnpm --filter movehat test:coverage`
- [x] `pnpm check:example`
- [x] `cd packages/docs && npx next build`
- [x] Empirical: `/tmp/fork-gap-demo/demo.mjs` → POST /v1/view returns 200 with real testnet supply
- [ ] After merge to `develop` → `main` batch: `curl -s -o /dev/null -w "%{http_code}\n" https://movehat.org/docs/guides/tutorial-fork-testing` returns 200 with the new view-fn step rendered

Closes #243
Tracks #238

### DIFF
--- a/ROADMAP.md
+++ b/ROADMAP.md
@@ -303,7 +303,8 @@ Follow-up issues filed during M4 (all upstream Movement CLI / SDK, deferred to M
 
 | Sub | Description | Issue | Status |
 |---|---|---|---|
-| M8.1 | Fork-mode tutorial + `networks-and-modes` accuracy fix | [#239](https://github.com/gilbertsahumada/movehat/issues/239) | in PR |
+| M8.0 | ForkServer `/v1/view` proxy + docs flip | [#243](https://github.com/gilbertsahumada/movehat/issues/243) | in PR |
+| M8.1 | Fork-mode tutorial + `networks-and-modes` accuracy fix | [#239](https://github.com/gilbertsahumada/movehat/issues/239) | ✅ shipped in PR #244 |
 | M8.2 | Deploy-to-live tutorial | [#240](https://github.com/gilbertsahumada/movehat/issues/240) | |
 | M8.3 | CI integration tutorial + reference workflow | [#241](https://github.com/gilbertsahumada/movehat/issues/241) | |
 | M8.4 | TypeDoc structural cleanup + MAINTENANCE.md + KPI2 report | [#242](https://github.com/gilbertsahumada/movehat/issues/242) | |
--- a/packages/docs/content/docs/guides/networks-and-modes.mdx
+++ b/packages/docs/content/docs/guides/networks-and-modes.mdx
@@ -11,7 +11,7 @@ Movehat exposes three `Harness` factories that target three different execution 
 | Your goal | Factory | Network | Writes gas? | Typical file |
 |---|---|---|---|---|
 | Fast iteration on a clean fresh chain (default for tests) | `createLocal()` | local node on `127.0.0.1:8080` | no (fake gas from faucet) | `tests/*.test.ts` |
-| Read real testnet / mainnet **account + resource** state | `createFork(network)` | account/resource reads passthrough; writes rejected | no | `tests/*.test.ts` |
+| Read real testnet / mainnet state (accounts, resources, view functions) | `createFork(network)` | reads proxied to upstream; transactions rejected | no | `tests/*.test.ts` |
 | Deploy or call entry functions on real testnet / mainnet | `createLive(network)` | live RPC, real submission | **yes — real MOVE** | `scripts/*.ts` |
 
 The rest of this page expands each row with code and explains the gotchas.
@@ -58,10 +58,12 @@ const account = await harness.runtime.aptos.getAccountInfo({
   accountAddress: "0x1",
 });
 
-// Reading a specific resource works
-const resource = await harness.runtime.aptos.getAccountResource({
-  accountAddress: "0x1",
-  resourceType: "0x1::account::Account",
+// Running a view function against real mainnet state works
+// (the fork server proxies POST /v1/view to upstream RPC)
+const [supply] = await harness.runViewFunction({
+  function: "0x1::coin::supply",
+  typeArguments: ["0x1::aptos_coin::AptosCoin"],
+  functionArguments: [],
 });
 
 // Deploy/upgrade/script attempts throw synchronously from the harness Proxy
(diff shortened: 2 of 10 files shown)


gilbertsahumada (commented):
## Self-review (§8)

### 🔴 Blockers

None.

### 🟡 Worth fixing / explicitly deferring

1. **Header passthrough is incomplete.** The proxy forwards the JSON body and sets `Content-Type: application/json` on the outgoing request, but does not forward request headers like `Accept` or `X-Aptos-Client`. For the standard JSON view-fn use case this is fine; for a caller that wants BCS-encoded view responses (`Accept: application/x-bcs`), the proxy currently silently returns JSON. Deferred: no current caller needs BCS view responses, and the SDK paths in the example all use JSON. Filing as a follow-up if needed.

2. **`MoveContract.call` still fails opaquely with HTTP 404 from the proxy.** The harness Proxy does not gate fork-mode `MoveContract.call` calls — they hit `/v1/transactions`, which still 404s (this PR deliberately scopes to view-fn only; writable-fork is #192). The error shape now is `{message: "Endpoint not found: /v1/transactions", error_code: "endpoint_not_found", ...}` which is structured but not as helpful as the synchronous harness-Proxy guards for `deployCodeObject` / `upgradeCodeObject` / `runMoveScript`. A ~5 LoC harness guard for `MoveContract.call` would improve the DX without expanding scope. Deferred: separate, optional small PR after M8 if useful.

3. **`server.ts` per-file coverage is at 56.95%** which is below the 80% gate that other critical modules hit, but `server.ts` is not in the critical-modules list in `vitest.config.ts` so the gate is not violated. The new `handleViewProxy` paths I added are covered by the 3 new vitest cases; the uncovered lines (376-385, 443-454) are CLI banner logging and shutdown handlers that don't fire during unit tests. Worth tracking as a separate coverage cleanup if `server.ts` ever gets added to the critical list.

### 🟢 Nits

- The body-size ceiling for view payloads is hard-coded to 1 MiB as a local constant inside `handleViewProxy`. If we ever need to tune it, it should move to a class field (matching the `MAX_VIEW_BODY_BYTES` pattern but exposed). Not urgent.
- `forwardView`'s "Fork not initialized" check throws the same string used elsewhere in `manager.ts`. Could share a constant but the four call sites that use this exact string are not worth a refactor today.

### Resolution

No 🟡 fixes applied in this PR. The three 🟡 items are all deferred:
- (1) header passthrough → file as separate issue if any caller needs BCS view responses.
- (2) `MoveContract.call` harness guard → already in the M8 "out of scope" list per the plan; small follow-up PR if useful.
- (3) `server.ts` coverage cleanup → separate effort, not blocking.

gilbertsahumada (commented):
## Resolution of §8 self-review yellow items (commit 625e26e)

🟡 (1) **MoveContract.call now guarded synchronously in fork mode** — `Harness.createFork` wraps `ctx.runtime.getContract` so contracts it returns are Proxy'd via `createForkContractProxy`. Calling `.call(...)` throws synchronously: `"Harness.createFork is read-only; MoveContract.call requires Harness.createLocal or createLive. Writable-fork support is tracked in #192."` `.view()` / `getModuleId()` pass through. Two vitest cases in `Harness.proxy.test.ts`.

🟡 (2) **Header passthrough implemented** — `MovementApiClient.view(payload, extraHeaders)` + `ForkManager.forwardView(payload, extraHeaders)` + a header whitelist in `handleViewProxy` (`Accept`, `X-Aptos-Client`). Hop-by-hop headers (`host`, `connection`, `content-length`) intentionally not forwarded. One vitest case verifies the forward shape.

🟡 (3) `server.ts` coverage cleanup — **still deferred**. Adding the new `handleViewProxy` branches lifted touched server.ts paths into coverage; the uncovered lines remain the same CLI banner + shutdown-handler paths that don't fire in unit tests. Not in the critical-modules gate; flagging as a separate effort if `server.ts` ever gets promoted.

## Verified

- `pnpm --filter movehat test` → **474/474** (was 471, +3 new)
- `pnpm --filter movehat test:coverage` → green (per-file gates unchanged)
- `pnpm check:example` → green
- `pnpm --filter movehat build` → clean
- Empirical: `/tmp/fork-gap-demo/demo.mjs` → `POST /v1/view` still returns 200 with real testnet supply

Ready for merge.

PR #258, opened 2026-05-21
test: arg-shape pass-through integration coverage (#98)

### LINKED ISSUE #98: [M1.7-adjacent] Integration test: arg-shape pass-through via real ChildProcessAdapter
## Why

The shell-escape regression caught in PR #97 (commit 511fd95) survived the unit test suite because tests inject a fake adapter that doesn't actually spawn a process. The bug only surfaced in `pnpm test:example` (~1m feedback loop) when the real Movement CLI received `'./move'` literally and failed to resolve the package.

A small integration test against `/bin/echo` would close this gap with ~20s feedback. The test exercises `defaultChildProcessAdapter.run()` end-to-end and asserts that args arrive un-mangled — catches shell-quote drift, arg-splitting bugs, env propagation issues, and any future "passes-through-the-redactor-but-mangled" surprises.

The new call-shape assertion in `__tests__/deployContract.test.ts` (commit 07cd30c) already covers the specific shell-escape regression at the unit level, but only by inspecting what `runtime.ts` *passes to* the adapter — not what the adapter *actually does* with it. The integration test plugs that remaining hole.

## Scope (deliberately minimal)

One test file: `packages/movehat/src/utils/__tests__/childProcessAdapter.integration.test.ts`.

```ts
import { describe, it, expect } from "vitest";
import { defaultChildProcessAdapter } from "../childProcessAdapter.js";

describe("defaultChildProcessAdapter — real-spawn arg-shape (integration)", () => {
  it("passes args verbatim with no shell quoting or splitting", async () => {
    const tricky = ["./path with space", "--flag=value-with-hyphens", "ed25519-priv-0xAA"];
    const result = await defaultChildProcessAdapter.run({
      command: "/bin/echo",
      args: tricky,
    });
    expect(result.exitCode).toBe(0);
    // /bin/echo prints args separated by single spaces; trailing newline.
    expect(result.stdout.trim()).toBe(tricky.join(" "));
  });

  it("preserves env when passed explicitly", async () => {
    const result = await defaultChildProcessAdapter.run({
      command: "/usr/bin/env",
      args: [],
      env: { ...process.env, MH_TEST_TOKEN: "alpha-beta_gamma" },
    });
    expect(result.exitCode).toBe(0);
    expect(result.stdout).toMatch(/^MH_TEST_TOKEN=alpha-beta_gamma$/m);
  });
});
```

That's it. No real Movement CLI, no fork server, no daemon — `/bin/echo` + `/usr/bin/env` are present on every supported platform (macOS, Linux). The first test would have caught the shell-escape regression in <100ms.

## Definition of Done

- [ ] `packages/movehat/src/utils/__tests__/childProcessAdapter.integration.test.ts` exists with the two tests above.
- [ ] Test file passes locally on macOS and Linux (CI matrix when CI reactivates per M4).
- [ ] No new dependencies. No new mocha/vitest config changes — file uses the existing vitest config.
- [ ] Unit suite count: 171 → 173.
- [ ] Document in `packages/movehat/CLAUDE.md` or equivalent that any new `ChildProcessAdapter` extension must include a parallel integration test against `/bin/echo` or `/bin/false` (one-line policy note).

## Why M1.7 and not standalone

M1.7 (#82) already touches `runCli.ts` + `childProcessAdapter.ts` for the strict-types audit (`any`, `!`, `noUncheckedIndexedAccess`). Adding two tests there is a marginal ~30 lines on top of an already-open work surface — opening a separate PR for this would be 1 commit, 1 file, less reviewable. If M1.7 ships independently, fold the tests in. If M1.7 sits idle, this issue can graduate to its own sub-PR.

## Refs

- Tracks #68 (M1 meta).
- Sibling of #82 (M1.7 strict-types audit).
- Regression motivating this issue: PR #97 commit 511fd95.

gilbertsahumada:
## Closes #98

Adds integration coverage for arg-shape pass-through via the real `defaultChildProcessAdapter`. The original bug (PR #97 / `./move` regression) survived the unit suite because injected fake adapters don't actually spawn a process.

### What the 6 new tests cover

Each uses Node binary as a deterministic spawn target. Helper asks the child to JSON-stringify its `argv.slice(1)`, then asserts the parsed array equals the input.

1. **path-with-slash literally** (`./move`) — the original regression
2. **embedded spaces** (`'hello world'`) — classic shell-split failure mode
3. **flag-style with `=`** (`--named-addresses counter=0x1`) — Movement CLI's actual usage
4. **shell metacharacters** (`$HOME`, `` `whoami` ``, `$(ls)`, `|`, `&&`) — proves we're using `spawn(cmd, args)` not `exec(cmdString)`
5. **empty-string** (`['first', '', 'third']`) — edge case
6. **unicode** (`español`, `日本語`, `🚀`) — encoding sanity

Note: helper inserts `--` after the inline `-e` script. Without it, Node consumes `--package-dir`-style args as Node flags and exits with code 9 (Invalid argument).

### Verification

- [x] Pre-flight: `pnpm --filter movehat exec vitest run src/utils/__tests__/childProcessAdapter.test.ts` → **25/25 pass**
- [x] Full suite: `pnpm --filter movehat test` → **513 pass** (was 507; +6)
- [x] Pre-push hook green (typecheck + build)
- [ ] CI on this PR
- [ ] §8 self-review before merge

### Why no Tier 2/3

Zero production code change. Test file only. Tier 2 (smoke) and Tier 3 (e2e) verify packaging/install which are unaffected. Marked N/A per CLAUDE.md §6.2/§6.3 rationale: change cannot affect the published artifact.

### Non-regression

Per user priority — verified the change introduces zero risk:
- No `src/**` files touched
- No exports changed
- Existing 507 tests all still pass

## Test plan

- [x] Unit tests (513 pass)
- [ ] CI matrix (Node 20/22)
- [ ] E2E (should pass — unrelated to test additions)

### DIFF
--- a/packages/movehat/src/utils/__tests__/childProcessAdapter.test.ts
+++ b/packages/movehat/src/utils/__tests__/childProcessAdapter.test.ts
@@ -288,4 +288,58 @@ describe('defaultChildProcessAdapter.run with inheritStdio', () => {
       setTimeoutSpy.mockRestore();
     }
   });
+
+  // Regression coverage for the arg-shape pass-through that fake-adapter unit
+  // tests cannot exercise. The original bug was './move' being eaten somewhere
+  // between Movehat and the Movement CLI; spawn(cmd, args) is the contract we
+  // need to keep honoring forever.
+  describe('arg-shape pass-through', () => {
+    async function spawnAndCaptureArgv(extraArgs: string[]): Promise<string[]> {
+      // `--` after the script tells Node to treat following args as positional
+      // (not Node flags), so '--package-dir' etc. survive intact in argv.
+      const result = await defaultChildProcessAdapter.run({
+        command: NODE,
+        args: [
+          '-e',
+          'process.stdout.write(JSON.stringify(process.argv.slice(1)))',
+          '--',
+          ...extraArgs,
+        ],
+      });
+      expect(result.exitCode).toBe(0);
+      return JSON.parse(result.stdout);
+    }
+
+    it('passes path-with-slash literally (the ./move regression)', async () => {
+      expect(await spawnAndCaptureArgv(['./move'])).toEqual(['./move']);
+    });
+
+    it('passes args with embedded spaces as single argv entries', async () => {
+      expect(await spawnAndCaptureArgv(['hello world', 'foo'])).toEqual(['hello world', 'foo']);
+    });
+
+    it('passes flag-style args with equals signs', async () => {
+      expect(
+        await spawnAndCaptureArgv(['--package-dir', '.', '--named-addresses', 'counter=0x1'])
+      ).toEqual(['--package-dir', '.', '--named-addresses', 'counter=0x1']);
+    });
+
+    it('passes shell metacharacters literally (no shell interpretation)', async () => {
+      expect(
+        await spawnAndCaptureArgv(['$HOME', '`whoami`', '$(ls)', '|', '&&'])
+      ).toEqual(['$HOME', '`whoami`', '$(ls)', '|', '&&']);
+    });
+
+    it('passes empty-string args as empty entries (not dropped)', async () => {
+      expect(await spawnAndCaptureArgv(['first', '', 'third'])).toEqual(['first', '', 'third']);
+    });
+
+    it('passes unicode args correctly', async () => {
+      expect(await spawnAndCaptureArgv(['español', '日本語', '🚀'])).toEqual([
+        'español',
+        '日本語',
+        '🚀',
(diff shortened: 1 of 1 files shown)


gilbertsahumada (commented):
§8 self-review: zero production code change, 54 LoC of test additions. Helper pattern matches existing tests in the same file (NODE binary + JSON.stringify(argv)). The 6 cases are exhaustive without redundancy. Local verification: 25/25 in file, 513/513 in full suite. No findings. Ready to merge.

PR #260, opened 2026-05-21
test: lock in fork address normalization consistency (#48)

### LINKED ISSUE #48: Tighten fork server address regex — minimum length
## Location
\`packages/movehat/src/fork/server.ts:137-149\`

## Problem
Regex allows \`0x[a-fA-F0-9]{1,64}\` — accepts addresses as short as 1 hex char. Aptos/Movement addresses are 32 bytes (64 hex). Code elsewhere normalizes to 64, so short matches skip through inconsistent code paths.

## Fix
Either normalize in the handler before dispatch or tighten the regex to the canonical \`{1,64}\` → fold into normalization utility. Document the expected range.

## Acceptance
- [ ] Consistent address handling across endpoints
- [ ] Tests covering short and padded addresses

gilbertsahumada:
## Closes #48

The audit-pass concern (#48) was that the permissive path regex `0x[a-fA-F0-9]{1,64}` would let short / padded / mixed-case forms reach downstream as different cache keys. **Investigation shows that's NOT actually happening today** — `ForkManager.getAccount` / `getResource` (`fork/manager.ts:117,142`) already call `normalizeAddress(input)` before storage lookup, so the cache is consistent regardless of input form.

### Why the regex stays permissive (intentional)

Movement framework addresses use short forms (`0x1` for framework, `0xa` for token, etc.) — tightening the minimum length would reject valid Movement addresses. The right design is permissive parsing + canonical normalization downstream, which is what we have.

### What this PR adds

4 regression tests locking in the consistency:

1. **short address** (`0x1`) reaches upstream with canonical padded form
2. **padded form** (`0x0..01`) reaches upstream with same canonical key
3. **mixed-case** (`0xABCdef`) is normalized to lowercase
4. **cache consistency**: a short request followed by a padded request for the same address hits the same on-disk cache entry (1 upstream call total, not 2)

Test #4 is the strongest proof — it shows the storage layer collapses both forms to the same canonical key.

### Verification

- [x] Pre-flight: `pnpm --filter movehat exec vitest run src/fork/__tests__/server.test.ts` → **15/15 pass**
- [x] Full suite: `pnpm --filter movehat test` → **517 pass** (was 513; +4)
- [x] Pre-push hook green
- [ ] CI on this PR

### Non-regression

Zero production code change. Test additions only. Failure mode: only the test would fail.

Per user priority: doesn't touch any `src/**` file.

## Test plan

- [x] Unit tests (517 pass)
- [ ] CI third-party gates (CodeRabbit / GitGuardian / Socket)

## Summary by CodeRabbit

* **Tests**
  * Added test suite validating address normalization consistency in the ForkServer's account endpoint. Ensures different address formats (short, padded, mixed-case) map to identical canonical forms and that cached results persist correctly across server restarts.

[Review Change Stack](https://app.coderabbit.ai/change-stack/gilbertsahumada/movehat/pull/260?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/packages/movehat/src/fork/__tests__/server.test.ts
+++ b/packages/movehat/src/fork/__tests__/server.test.ts
@@ -271,3 +271,99 @@ describe("ForkServer — POST /v1/view proxy", () => {
     expect(body.error_code).toBe("malformed_path");
   });
 });
+
+// Regression coverage for #48 — the path regex deliberately accepts
+// `0x[a-fA-F0-9]{1,64}` because Movement framework addresses use short forms
+// (0x1, 0xa, ...). What matters is that downstream consistently normalizes,
+// so short / padded / mixed-case inputs all collapse to the same canonical key.
+describe("ForkServer — GET /v1/accounts/:address normalization consistency (#48)", () => {
+  let forkDir: string;
+  let server: ForkServer | null = null;
+
+  beforeEach(() => {
+    forkDir = makeForkDir();
+    fakeApi.getAccount.mockReset();
+  });
+
+  afterEach(async () => {
+    if (server) {
+      await server.stop();
+      server = null;
+    }
+    rmSync(forkDir, { recursive: true, force: true });
+  });
+
+  async function startAndGetAccount(path: string): Promise<Response> {
+    server = new ForkServer(forkDir, 0);
+    await server.start();
+    const internal = (server as unknown as { server: { address(): AddressInfo } }).server;
+    const port = internal.address().port;
+    return fetch(`http://127.0.0.1:${port}${path}`);
+  }
+
+  const PADDED_ONE = "0x" + "0".repeat(63) + "1";
+
+  it("short address (0x1) reaches upstream with the canonical padded form", async () => {
+    fakeApi.getAccount.mockResolvedValue({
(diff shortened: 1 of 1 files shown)


gilbertsahumada:
@coderabbitai review

gilbertsahumada (commented):
§8 self-review:

🔴 Blockers: none.
🟡 Worth fixing: none.
🟢 Nits:
- The 4 tests use a shared `PADDED_ONE` constant for clarity. Acceptable since it's read-only.
- Helper `startAndGetAccount` mirrors the existing `startAndFetch` shape from the same file — consistent.

**Investigation finding logged in PR body**: the regex 'tightening' the issue requested isn't actually needed — `ForkManager.getAccount/getResource` already normalize via `utils/address.ts:normalizeAddress` before storage lookup. PR locks in that consistent behavior with regression tests instead of changing the regex.

**CodeRabbit**: skipped per repo policy (auto-reviews disabled on non-default branches). Will get reviewed at next develop→main batch like the M8 PRs did.

**Non-regression**: zero `src/**` change. Tests only. 513 → 517 unit tests, all green.

Ready to merge.

PR #261, opened 2026-05-21
fix(parser): drop fragile fallback regex in parseTxHash (#51)

### LINKED ISSUE #51: [bug] Fragile tx-hash parsing from Movement CLI stdout
## Location
\`packages/movehat/src/runtime.ts:288-301\`

## Problem
Primary regex matches \`transaction/txn/hash: 0x…64hex\` — good. Fallback regex \`\\b(0x[a-fA-F0-9]{64})\\b\` matches the **first** 64-char hex string in stdout. If Movement CLI ever prints another 64-hex token first (e.g. a padded module address, a state root), we cache a wrong \`txHash\` silently.

## Fix
- Drop the fallback; if the contextual regex fails, throw / warn \`txHash=undefined\` explicitly.
- Prefer Movement CLI \`--json\` / \`--output-format json\` output if available and parse a structured response.

## Acceptance
- [ ] Only contextual match accepted
- [ ] Unit tests cover: success, no hash in output, multiple hashes in output

gilbertsahumada:
## Closes #51

The previous `parseTxHash` fell back to matching the FIRST 64-char hex literal when the contextual pattern (`transaction|txn|hash: 0x…`) missed. This silently returned the WRONG hash whenever Movement CLI printed another 64-hex token before the actual transaction hash (e.g. a padded module address, a state root).

### What changes in production

`packages/movehat/src/utils/parseCliOutput.ts`:
- Removed the loose `\b(0x[a-fA-F0-9]{64})\b` fallback
- Added `logger.warning` when no contextual match is found
- Still returns `undefined` — caller behavior is unchanged

### Caller verification (no behavior regression)

All 3 callers already tolerated `undefined` from the parser:
- `core/Publisher.ts:245` — assigns to optional `deployment.txHash` field, silent acceptance
- `harness/codeObject.ts:290` — same pattern
- `harness/script.ts:154` — explicitly `if (!txHash) throw new Error(...)` with clear message

The only behavior change: when context-match fails and there's a hex blob in stdout, today's code returns a wrong-hash silently; after this fix, returns `undefined` with a logger warning. Publisher/codeObject record a deployment with no txHash (visible). Script throws (loud) — same as today.

### New regression tests

`packages/movehat/src/utils/__tests__/parseCliOutput.test.ts` (NEW, 6 cases):

1. context-bearing pattern extracts correctly
2. `txn hash:` and `hash:` variants extract correctly
3. **the #51 trap**: loose 64-hex literal returns undefined + warns (would have caught the silent wrong-hash regression)
4. when BOTH trap-hash AND context-hash appear, the context-bearing one wins (NOT the first match — this was the original bug)
5. empty stdout returns undefined + warns
6. case-insensitive on the keyword

### Verification

- [x] Pre-flight: `pnpm --filter movehat exec vitest run src/utils/__tests__/parseCliOutput.test.ts` → **6/6 pass**
- [x] Full suite: `pnpm --filter movehat test` → **523 pass** (517 → +6)
- [x] Pre-push hook green
- [ ] CI on this PR

### Non-regression (user priority)

- Caller tolerance verified (see "Caller verification" above)
- Failure mode shifts from SILENT WRONG-HASH to LOUD WARNING — strictly better
- Script-path users get a clear thrown error (same as today)
- Publisher/codeObject deployment records may have missing `txHash` in the rare case context-match fails — but the deployment IS on chain, this is a bookkeeping degradation not a deploy failure

## Test plan

- [x] Unit suite (523)
- [x] Caller tolerance audit
- [ ] CI third-party gates

### DIFF
--- a/packages/movehat/src/utils/__tests__/parseCliOutput.test.ts
+++ b/packages/movehat/src/utils/__tests__/parseCliOutput.test.ts
@@ -1,0 +1,63 @@
+import { describe, it, expect, vi, beforeEach, afterEach } from 'vitest';
+import { parseTxHash } from '../parseCliOutput.js';
+
+// Real-world-ish hashes — distinct values so the "wrong-fallback" trap
+// is visible if anyone re-introduces the loose regex.
+const REAL_TX_HASH = '0x' + 'a'.repeat(64);
+const TRAP_HASH = '0x' + 'b'.repeat(64); // padded module address, state root, etc.
+
+describe('parseTxHash (regression: #51)', () => {
+  let warnSpy: ReturnType<typeof vi.spyOn>;
+
+  beforeEach(() => {
+    // Logger uses console.warn under the hood (see ui/logger.ts:139).
+    // Spy at the console level so the assertion is decoupled from logger
+    // formatting choices.
+    warnSpy = vi.spyOn(console, 'warn').mockImplementation(() => undefined);
+  });
+
+  afterEach(() => {
+    warnSpy.mockRestore();
+  });
+
+  it('extracts hash from the context-bearing pattern (transaction hash: 0x…)', () => {
+    const stdout = `Some prelude line\nTransaction hash: ${REAL_TX_HASH}\nMore output`;
+    expect(parseTxHash(stdout)).toBe(REAL_TX_HASH);
+    expect(warnSpy).not.toHaveBeenCalled();
+  });
+
+  it('extracts hash from "txn hash:" and "hash:" variants too', () => {
+    expect(parseTxHash(`txn hash: ${REAL_TX_HASH}`)).toBe(REAL_TX_HASH);
+    expect(parseTxHash(`hash: ${REAL_TX_HASH}`)).toBe(REAL_TX_HASH);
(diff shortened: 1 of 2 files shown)


gilbertsahumada (commented):
§8 self-review:

🔴 Blockers: none.
🟡 Worth fixing: none.
🟢 Nits:
- Used `afterEach(() => warnSpy.mockRestore())` to avoid spy stack across tests (caught locally during dev — initial run had count leaks).
- Could also have used `vi.restoreAllMocks()` for broader hygiene; restored just the specific spy for clarity.

**Caller verification (matters most for this PR)** — confirmed in PR body:
- Publisher (line 245) → optional `deployment.txHash` field, undefined OK
- codeObject (line 290) → same pattern, undefined OK
- script (line 154) → already throws on undefined, behavior unchanged

**Behavior shift**: SILENT WRONG-HASH → LOUD WARNING. Strictly safer.

**CodeRabbit**: skipped per repo policy (auto-reviews disabled on non-default branches). Will get reviewed at next develop→main batch.

Ready to merge.

PR #262, opened 2026-05-21
fix(security): require URL allowlist before auto-injecting test key (#40)

### LINKED ISSUE #40: [security] Test key auto-applied based on network *name* — spoofable
## Severity
Critical (low probability, high impact footgun)

## Location
`packages/movehat/src/core/config.ts:146`

## Problem
The check `if (selectedNetwork === 'testnet' || selectedNetwork === 'local')` auto-injects the deterministic test key `0x…01`. A user who defines a network named `testnet` in `movehat.config.ts` but points it at a production-like RPC would inherit the test key automatically.

## Fix
- Combine name + validate chainId / URL against an allowlist of known Movement test endpoints, OR
- Require an explicit flag (e.g. `--use-test-account`) before injecting the deterministic key, OR
- Only auto-inject when the URL host matches `*.movementnetwork.xyz` testnet / `localhost` / `127.0.0.1`.

## Acceptance
- [ ] Can't get test key on a custom-named network without opt-in
- [ ] Test covers a user-named `testnet` pointing elsewhere

gilbertsahumada:
## Closes #40

The deterministic test key (`0x0..01`) was previously injected whenever the network NAME matched `testnet` or `local` — regardless of what URL the user pointed that network at. A user who defined:

```ts
{ networks: { testnet: { url: 'https://prod.example.com/v1', ... } } }
```

…silently got the well-known test key with a production endpoint. Critical (low-probability, high-impact) footgun, especially with `setupTestFixture` flows that don't surface the key being used.

### Fix

Gate the test-key injection on **both** network name AND URL host being in an allowlist:

```
TEST_ENDPOINT_HOSTS = { testnet.movementnetwork.xyz, localhost, 127.0.0.1, ::1 }
```

When name matches but URL doesn't:
1. Log a `logger.warning` explaining why injection was skipped
2. Fall through to the standard "no accounts configured" throw with actionable guidance (PRIVATE_KEY env, explicit `accounts` config, or rename)

### Failure mode shift

| Before | After |
|---|---|
| SILENT test-key injection with arbitrary URL | LOUD warning + explicit "no accounts" error unless URL matches allowlist |

Strictly safer. No more silent test-key on production.

### Backward-compat (non-regression check)

- Users with literal `testnet` / `local` pointing at the canonical URLs → **IDENTICAL behavior** (test key injected as before)
- Users with explicit `accounts` in their config → **unaffected** (the auto-inject branch only fires when `accounts` is empty)
- Users with `PRIVATE_KEY` env var → **unaffected** (env path takes precedence)
- Users who renamed `testnet` to point at custom URL → **now get a clear warning + error** instead of silently using the wrong key

### Phase 3 deferred to 0.3.0

This PR ships Phases 1+2 (gate + warn). Phase 3 (hard-fail with breaking error message instead of warn-and-skip) is deferred to 0.3.0 per MAINTENANCE.md deprecation policy.

### Verification

- [x] Pre-flight: `pnpm --filter movehat exec vitest run src/core/__tests__/config.test.ts` → **30/30 pass** (was 24; +6 new)
- [x] Full suite: `pnpm --filter movehat test` → **529 pass** (was 523; +6)
- [x] Pre-push hook green
- [ ] CI on this PR

### New tests (6 cases)

1. canonical Movement testnet URL → injects test key
2. localhost URL with `local` network → injects test key
3. **THE #40 TRAP**: `testnet` name + production URL → throws + warn
4. `local` name + remote URL → throws
5. `127.0.0.1` is recognized as test endpoint
6. malformed URL (no `://`) → throws (no injection)

## Test plan

- [x] Unit suite (529)
- [x] Backward-compat audit
- [ ] CI third-party gates

### DIFF
--- a/packages/movehat/src/core/__tests__/config.test.ts
+++ b/packages/movehat/src/core/__tests__/config.test.ts
@@ -374,4 +374,84 @@ describe("resolveNetworkConfig", () => {
     const resolved = await resolveNetworkConfig(user, "a");
     expect(resolved.profile).toBe("custom-profile");
   });
+
+  // Regression coverage for #40 — auto-injection of the deterministic test
+  // key (0x0..01) was previously gated only on the network NAME being
+  // 'testnet' / 'local'. A user-named 'testnet' pointing at a production
+  // URL silently inherited the weak key. Now name AND URL host must match.
+  describe("test-key auto-injection gate (#40)", () => {
+    let warnSpy: ReturnType<typeof vi.spyOn>;
+
+    beforeEach(() => {
+      warnSpy = vi.spyOn(console, "warn").mockImplementation(() => undefined);
+    });
+
+    afterEach(() => {
+      warnSpy.mockRestore();
+    });
+
+    it("injects test key for 'testnet' network with canonical Movement testnet URL", async () => {
+      const user = baseUserConfig({
+        testnet: {
+          url: "https://testnet.movementnetwork.xyz/v1",
+          chainId: "testnet",
+        },
+      });
+      const resolved = await resolveNetworkConfig(user, "testnet");
+      expect(resolved.privateKey).toBe(
+        "0x0000000000000000000000000000000000000000000000000000000000000001"
+      );
+    });
+
+    it("injects test key for 'local' network with localhost URL", async () => {
+      const user = baseUserConfig({
+        local: { url: "http://localhost:8080/v1", chainId: "local" },
+      });
+      const resolved = await resolveNetworkConfig(user, "local");
+      expect(resolved.privateKey).toBe(
+        "0x0000000000000000000000000000000000000000000000000000000000000001"
+      );
+    });
+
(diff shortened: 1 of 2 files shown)


gilbertsahumada (commented):
§8 self-review:

🔴 None.
🟡 None.
🟢 Nits:
- TEST_ENDPOINT_HOSTS as a Set + helper function isolates the allowlist from the gate logic — easy to extend.
- The fall-through pattern (warn + skip + standard throw) reuses the existing error message for consistency.

**Behavior shift verification**:
- Canonical testnet URL → injects (same as before) ✓
- Custom 'testnet' name + prod URL → previously SILENT, now WARNS + THROWS with actionable guidance
- Explicit accounts config → unaffected ✓
- PRIVATE_KEY env path → unaffected ✓

**Backward-compat audit**: did not find any caller / test that relied on the spoofable behavior. The existing 24 tests still pass unchanged.

**CodeRabbit**: skipped per repo policy (auto-reviews disabled on develop-targets). Will get reviewed at next develop→main batch.

Ready to merge.

PR #263, opened 2026-05-21
fix(security): prefer movehat-bundled tsx over cwd resolution (#52)

### LINKED ISSUE #52: [security] tsx resolved from cwd first — spoofable in untrusted dirs
## Location
\`packages/movehat/src/commands/run.ts:49\`

## Problem
\`require.resolve('tsx', { paths: [process.cwd()] })\` tries the cwd's \`node_modules\` before the movehat package itself. In an untrusted project dir, a malicious \`node_modules/tsx/dist/cli.mjs\` would execute.

## Risk
Low — users generally trust cwd they run tools in. But the bundler/package-manager community has been burned by supply-chain attacks in \`node_modules\`.

## Fix options
- Document that \`movehat run\` should only be used in trusted dirs.
- Or resolve from the movehat package first, fall back to cwd only with an explicit flag.
- Or pin \`tsx\` to movehat's own bundled copy and drop the cwd lookup entirely.

## Acceptance
- [ ] Decision documented (prefer movehat-bundled tsx by default)
- [ ] README warns of the behavior

gilbertsahumada:
## Closes #52

`movehat run <script>` previously resolved tsx from `process.cwd()` FIRST and fell back to movehat's bundled tsx only on failure. A malicious `node_modules/tsx/dist/cli.mjs` in an untrusted project directory would silently execute. Supply-chain class of attack.

### Fix

Bundled-first by default, cwd as opt-in:

1. Extracted resolution into exported `resolveTsxCliPath(opts?: { preferCwd?: boolean })` helper (now unit-testable)
2. Default lookup order: `[__dirname, process.cwd()]` (bundled first)
3. With `MOVEHAT_TSX_FROM_CWD=1` env var or `preferCwd: true`: `[process.cwd(), __dirname]` (cwd first)
4. When the env var is set, `logger.warning` fires so the choice is visible in the output

### Failure mode shift

| Before | After |
|---|---|
| Cwd's tsx loaded silently (malicious or not) | Bundled tsx loaded silently; cwd opt-in is loud |

### Backward-compat (non-regression check)

- `packages/movehat/package.json` declares `"tsx": "^4.7.0"` → bundled tsx is **always present** in a normal install
- For the 99% case (movehat installed via npm/pnpm, no custom tsx pinning), bundled-first succeeds → cwd lookup never reached → **identical observable behavior**
- Users who pinned a different tsx and need it: set `MOVEHAT_TSX_FROM_CWD=1` (and accept the security warning)
- `tsx` is a CLI runner (not a library) — different versions are interface-compatible for `tsx <file>` invocation

### New tests (6 cases)

In `run.test.ts`:

1. `resolveTsxCliPath()` returns a path ending in `dist/cli.mjs` under default
2. `resolveTsxCliPath({ preferCwd: false })` (explicit default) returns valid path
3. `resolveTsxCliPath({ preferCwd: true })` returns valid path (workspace has tsx via both routes)
4. Env var `MOVEHAT_TSX_FROM_CWD=1` is respected by the orchestrator
5. `runCommand` emits warning when env var set
6. `runCommand` does NOT emit warning under default behavior

Filesystem-controlled order assertions (proving bundled is loaded vs cwd version) are out of scope — see the pre-existing `tsx not found` test comment in the same file. The observable behavior (path returned, env var honored, warning fired) is the testable surface.

### Verification

- [x] Pre-flight: `pnpm --filter movehat exec vitest run src/commands/__tests__/run.test.ts` → **17/17 pass** (was 11; +6)
- [x] Full suite: `pnpm --filter movehat test` → **535 pass** (was 529; +6)
- [x] Pre-push hook green
- [ ] CI on this PR

## Test plan

- [x] Unit suite (535)
- [x] Backward-compat audit
- [ ] CI third-party gates

### DIFF
--- a/packages/movehat/src/commands/__tests__/run.test.ts
+++ b/packages/movehat/src/commands/__tests__/run.test.ts
@@ -13,7 +13,7 @@ vi.mock("../../utils/runCli.js", () => ({
 }));
 
 // Static import after mock declaration — vi hoists vi.mock.
-const { default: runCommand } = await import("../run.js");
+const { default: runCommand, resolveTsxCliPath } = await import("../run.js");
 
 /**
  * Direct coverage for the signal-forwarding branch added to fix the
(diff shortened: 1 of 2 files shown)


gilbertsahumada (commented):
§8 self-review:

🔴 None.
🟡 None.
🟢 Nits:
- The 6 tests focus on observable behavior (path returned, env honored, warn fired) — filesystem-order assertions deferred per the pre-existing 'tsx not found' note in the same file.
- The orchestrator warn-emission tests use try/catch around runCommand because process.exit propagation may throw; the warning always fires before any exit path so this is fine.

**Backward-compat verification**: bundled tsx is a declared dep in movehat's package.json → always present in normal installs. For the 99% case, lookup order change is invisible (same tsx, same code, same behavior).

**CodeRabbit**: skipped per repo policy (develop-target). Will get reviewed at next develop→main batch.

Ready to merge.

PR #264, opened 2026-05-21
fix(config): dedupe concurrent loadUserConfig calls (#47)

### LINKED ISSUE #47: [bug] tsx loader register/unregister is not reentrant in loadUserConfig
## Location
\`packages/movehat/src/core/config.ts:42-50\`

## Problem
\`loadUserConfig\` calls \`register()\` then \`unregister()\` around dynamic \`import()\`. Two concurrent \`loadUserConfig\` calls would see the first \`unregister()\` remove the tsx loader while the second is still importing → load error.

## Fix
- Ref-count the register/unregister (only unregister when the last caller finishes), OR
- Cache the loaded config module by \`path + mtime\` so we only register once per process, OR
- Serialize all config loads with an in-process mutex.

## Acceptance
- [ ] Two concurrent \`loadUserConfig\` calls both succeed
- [ ] Test with \`Promise.all([loadUserConfig(), loadUserConfig()])\`

gilbertsahumada:
## Closes #47

Two concurrent `loadUserConfig()` calls hitting a cold cache for the same `.ts` config previously both invoked tsx's `register()`/`unregister()` cycle. If the first caller's `await import()` finished before the second's, `unregister()` would remove the tsx loader mid-flight for the second caller — load error.

### Fix

**In-flight Promise dedup.** A new module-scope `inFlightLoads: Map<string, Promise<MovehatUserConfig>>` tracks active load promises by `configPath`. Concurrent callers share the SAME promise instead of starting another load. After the promise settles, the entry is cleaned up so a later mtime change can trigger a fresh cold load.

Structural change: the cold-load body extracted into `doLoadConfig` helper. All callers (the originator and dedup'd ones) flow through the same `await` / outer try-catch so they observe the same wrapped error on failure.

### Race scenario closed

```
BEFORE (race):
T0 A: register() → loader_A
T0 B: register() → loader_B  (or stacks on top of loader_A — tsx-impl-dependent)
T1 A: import() returns; unregister() removes loader_A
T2 B: import() still in flight, loader chain shifted → potential failure

AFTER (dedup):
T0 A: inFlight.set(loadPromise); register() → loader (once)
T0 B: inFlight.get() returns A's loadPromise; B awaits it
T1 A: import() returns; unregister(); cleanup
T2 B: receives same module from A's load — no second register/unregister cycle
```

### New tests (3 cases in config.test.ts)

Note: tests use `.js` fixtures because vitest's vite-node loader collides with tsx's `register()` in-process (`Invalid loader value` error). The dedup mechanism lives BEFORE the .ts/.js branch in `doLoadConfig` — proving dedup with `.js` proves it for `.ts` in production where vitest is not in the loader chain.

1. **Promise.all returns same object** (3-way) — proves only one load ran
2. **Rejection propagates equivalently** to all concurrent callers — proves the dedup'd path produces the same wrapped error as the originating path
3. **Sequential call after concurrent burst hits cache normally** — proves the in-flight cleanup runs and doesn't leak

### Risk profile

- **Backward-compat**: sequential calls (the dominant case) flow through the same code path as today
- **Concurrent burst**: today multiple register/unregister cycles; after, one register/unregister + same module result. Observable behavior unchanged for the caller; race window closes
- **Cleanup**: `void loadPromise.finally(cleanup).catch(() => undefined)` ensures the cleanup-chain rejection (when load fails) is swallowed — original error still propagates to awaiters via `return await loadPromise`

### Verification

- [x] Pre-flight: `pnpm --filter movehat exec vitest run src/core/__tests__/config.test.ts` → **33/33 pass** (was 30; +3)
- [x] Full suite: `pnpm --filter movehat test` → **538 pass** (was 535; +3)
- [x] Pre-push hook green
- [ ] CI on this PR

## Test plan

- [x] Unit suite (538)
- [x] Risk audit (sequential, concurrent, rejection paths)
- [ ] CI third-party gates

### DIFF
--- a/packages/movehat/src/core/__tests__/config.test.ts
+++ b/packages/movehat/src/core/__tests__/config.test.ts
@@ -137,6 +137,72 @@ describe("loadUserConfig — mtime cache (#81, #62)", () => {
     );
     await expect(loadUserConfig()).rejects.toThrow(/No networks defined/);
   });
+
+  // Regression coverage for #47 — concurrent cold-cache loads of the
+  // same config previously raced on tsx's register/unregister cycle
+  // (for .ts files). The in-flight dedup map makes concurrent callers
+  // share one Promise regardless of extension.
+  //
+  // Note: tests use .js fixtures because vitest's vite-node loader
+  // collides with tsx's `register()` in-process (`Invalid loader value`
+  // error). The dedup mechanism lives BEFORE the .ts/.js branch in
+  // doLoadConfig — proving dedup with .js proves it for .ts in
+  // production where vitest is not in the loader chain.
+  describe("concurrent load dedup (#47)", () => {
+    it("Promise.all on cold cache returns the same loaded object", async () => {
+      writeFileSync(join(tmpCwd, "movehat.config.js"), CONFIG_A);
+
+      const [a, b, c] = await Promise.all([
+        loadUserConfig(),
+        loadUserConfig(),
+        loadUserConfig(),
+      ]);
+
+      // Reference equality proves all three callers received the SAME
+      // resolved module — only one load actually ran.
+      expect(a).toBe(b);
+      expect(b).toBe(c);
+      expect(a.defaultNetwork).toBe("testnet");
+    });
+
+    it("Promise.all rejection propagates the same error to all callers", async () => {
+      // Config without `networks` triggers the validation throw inside
+      // doLoadConfig, exercising the rejection path of the in-flight
+      // promise.
+      writeFileSync(
+        join(tmpCwd, "movehat.config.js"),
+        `export default { networks: {} };
+`
+      );
+
+      const results = await Promise.allSettled([
+        loadUserConfig(),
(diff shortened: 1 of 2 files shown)


gilbertsahumada (commented):
§8 self-review:

🔴 None.
🟡 None.
🟢 Nits:
- The .catch(() => undefined) at the end of the cleanup chain is non-obvious but commented. Without it, a load rejection becomes an unhandled rejection (verified via vitest's 'Unhandled Rejection' surface during dev).
- toBe(reference equality) for rejection comparison is intentionally replaced with toBe(String(reason)) since each outer catch creates a new Error instance — content equality is the right invariant.
- Tests use .js fixtures because vitest's vite-node loader collides with tsx register in-process. Documented in the describe comment.

**Backward-compat**: sequential path identical, concurrent burst converges on the same module result. No observable user behavior change beyond closing the race window.

**CodeRabbit**: skipped per repo policy (develop-target).

Ready to merge.

PR #265, opened 2026-05-21
0.2.6 batch: 7 audit-pass items (tests + security hardening)

gilbertsahumada:
## Batch develop → main for the 0.2.6 release cycle

7 audit-pass PRs + 1 sync (SECURITY_AUDIT removal mirror to develop). Total 11 files changed, 673 insertions / 102 deletions.

### Sub-PRs in this batch

| Sub-PR | Issue | Risk | Scope |
|---|---|---|---|
| #258 | #98 | zero | arg-shape pass-through integration test |
| #259 | #214 | zero | pin GitHub Actions to commit SHAs |
| #260 | #48 | low | fork address normalization consistency tests |
| #261 | #51 | low | drop fragile tx-hash fallback in `parseTxHash` |
| #262 | #40 | medium | URL allowlist gate before auto-injecting test key |
| #263 | #52 | medium | prefer movehat-bundled tsx (cwd opt-in via `MOVEHAT_TSX_FROM_CWD=1`) |
| #264 | #47 | medium | dedupe concurrent `loadUserConfig` calls |
| #257 | — | — | sync: SECURITY_AUDIT removal mirror to develop |

### Tests gained

507 → 538 unit tests (+31) across the 7 issue PRs. All run green locally via `pnpm --filter movehat test`.

### Verification per CLAUDE.md §6

- [x] Tier 1: `pnpm typecheck:example` + `pnpm --filter movehat build` (auto via pre-push on each sub-PR)
- [x] Unit + coverage: 538/538 passing
- [x] Sub-PRs each carried §8 self-review
- [ ] CI on this batch PR (full matrix + E2E)
- [ ] Post-merge: cut 0.2.6 release branch
- [ ] Post-release: sync release commit back to develop

### Non-regression summary

User priority for the entire audit-pass cycle was "no romper nada que funciona". Verified per PR:

- All 7 sub-PRs added regression tests proving the new behavior
- Backward-compat audit on each medium-risk PR (`#40`, `#52`, `#47`) documented in PR body
- Failure modes shifted from SILENT (wrong-hash, wrong-key, race) to LOUD (warn, throw, dedup) — strictly safer

### Issues auto-closing on merge

GitHub auto-close fires when this batch lands on `main`:
- #98 #214 #48 #51 #40 #52 #47

(All were manually closed earlier when their sub-PRs merged to develop per §10; this just re-confirms.)

## Test plan

- [x] Sub-PR sequence (7 green merges to develop)
[...]

PR #266, opened 2026-05-21
fix(security): address CodeRabbit findings on #265 batch

gilbertsahumada:
Two 🟡 findings from CodeRabbit's review of the 0.2.6 batch PR (#265). Both fixed here, in develop, so #265 auto-refreshes.

## Fixes

1. **docs-deploy.yml** — add `persist-credentials: false` on the checkout step. Workflow uses OIDC for Pages upload + never `git push`s, so GITHUB_TOKEN persistence is unnecessary token exposure. (zizmor 'artipacked')

2. **config.ts** — sanitize URL in the test-key-gate warning. Previously logged raw `networkConfig.url`, leaking userinfo (`user:pass@`) or query-string credentials (`?apiKey=…`) into terminal/CI logs. New `sanitizeUrlForLog` helper returns `protocol://host/pathname` only.

New test asserts a URL with userinfo + apiKey gets sanitized in the warn output (538 → 539 tests).

Pre-push hook green. Closes the 2 actionable CR comments on #265.

### DIFF
--- a/.github/workflows/docs-deploy.yml
+++ b/.github/workflows/docs-deploy.yml
@@ -24,6 +24,11 @@ jobs:
 
     steps:
       - uses: actions/checkout@34e114876b0b11c390a56381ad16ebd13914f8d5  # v4.3.1
+        with:
+          # docs-deploy never pushes back to the repo (Pages upload uses
+          # OIDC, not git push), so persisting GITHUB_TOKEN is unnecessary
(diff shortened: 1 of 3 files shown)


gilbertsahumada (commented):
§8 self-review: 2 CR findings addressed. 44 LoC swap (1 line + comment for persist-credentials, helper + use site + new test for URL sanitization). Test coverage on the sanitization path proves userinfo + apiKey are stripped. Ready to merge to develop so #265 batch refreshes.

Issue #54, opened 2026-04-25
[refactor] Replace dynamic await import() in deployContract with static imports


gilbertsahumada:
## Location
\`packages/movehat/src/runtime.ts\` — inside \`deployContract\`

## Problem
10+ \`await import(...)\` calls for \`child_process\`, \`util\`, \`fs\`, \`path\`, \`os\`, \`js-yaml\`, \`./core/shell.js\`, \`./commands/compile.js\` inside the hot path. This pattern was likely adopted to keep startup lean, but the imports execute on every deploy, add overhead, and obscure dependencies from static analysis.

## Fix
Move to static top-level \`import\`. Only \`tsx/esm/api\` (used in \`config.ts\`) genuinely needs to be dynamic.

## Acceptance
- [ ] All stdlib imports static at file top
- [ ] Tree-shaking / type-checking unaffected

gilbertsahumada:
Already resolved in commit `930442d` ("refactor(core): extract Publisher.ts scaffold", May 13, 2026) when the deploy logic was moved out of `runtime.ts` into `packages/movehat/src/core/Publisher.ts`.

The dynamic `await import()` calls for `fs` / `path` / `os` / `js-yaml` / `./core/shell.js` / `./commands/compile.js` cited in the issue are now static imports at the top of `Publisher.ts`. `tsx/esm/api` stayed dynamic in `config.ts` as the issue explicitly allowed.

Verified: `grep -n 'await import' packages/movehat/src/runtime.ts` returns no matches today.

Closing — durable record is the commit hash + this grep.

PR #273, opened 2026-05-21
test(deployContract): add lifecycle + idempotency unit tests

### LINKED ISSUE #61: [testing] Add unit tests for runtime.deployContract
## Problem
\`runtime.ts\` \`deployContract\` is the most security- and correctness-critical function in the project — and it has zero direct unit tests. It mutates shared system state, parses CLI output, and recovers from errors.

## Test cases to add (mocking \`exec\` / \`fs\` with \`memfs\` already in devDeps)
- Happy path: builds, publishes, parses tx hash, saves deployment.
- Already-deployed → throws \`ModuleAlreadyDeployedError\`.
- Force-redeploy via \`MH_CLI_REDEPLOY=true\`.
- Build failure → no config write, no Move.toml mutation persisted.
- Publish failure → original config and Move.toml restored.
- SIGINT during publish → cleanup runs, no key on disk (paired with #36).
- Concurrent deploys (paired with #37).

## Acceptance
- [ ] \`runtime.test.ts\` covers all scenarios
- [ ] Coverage for \`runtime.ts\` ≥ 80%

gilbertsahumada:
## Summary

Adds 4 unit tests to `packages/movehat/src/__tests__/deployContract.test.ts` covering the deploy lifecycle paths that lacked direct coverage after the Publisher.ts extraction (#54 / 930442d).

## Why now

#61 was filed before the Publisher.ts extraction; its acceptance asks for ≥80% coverage on "runtime.ts" but the deploy logic now lives in `src/core/Publisher.ts`. `runtime.ts` itself is at 100% (thin wrapper that delegates), but `Publisher.ts` was at **66% lines / 44% branches** — the actual gap behind the issue's intent.

## Test scenarios added

| # | Test | Why it matters |
|---|---|---|
| 1 | happy path — asserts full deployment-record shape on disk | Locks the persisted JSON shape so a refactor that breaks `saveDeployment` writes a wrong field is caught here, not in user repos |
| 2 | `ModuleAlreadyDeployedError` early-exit | Asserts the guard fires before any `runCli` call (0 adapter calls), preventing accidental redeploys |
| 3 | `MH_CLI_REDEPLOY=true` bypass | Same precondition as #2 but with the flag set; asserts deploy proceeds and overwrites the stale on-disk record |
| 4 | build failure aborts before publish | exitCode 1 from build → `CliExecutionError`, publish never invoked, no deployment file written |

## Coverage delta on `src/core/Publisher.ts`

| Metric | Before | After | Δ |
|---|---|---|---|
| Lines | 66.17% | **83.82%** | +17.65 |
| Statements | 63.38% | **80.28%** | +16.90 |
| Branches | 44.18% | 58.13% | +13.95 |
| Functions | 83.33% | 83.33% | — |

Both lines and statements now ≥80%. Branches still <70 because the PostPublishError + non-CLI catch-block branches are out of scope here (would require mocking `saveDeployment` to throw — a different test surface). No per-file branch gate is set for Publisher.ts.

`runtime.ts` already at 100% — unchanged.

## What's already covered (no duplication)

The existing 5 tests in this file cover:
- SIGINT cleanup (covers cross-issue #36)
- Concurrent isolation (covers cross-issue #37)
- Stderr redaction (leak path #2)
- Console-error redaction on successful-but-noisy publish (leak path #1)
- Move.toml non-mutation (#38)

The 4 new tests are additive and don't touch these scenarios.

## Test plan

- [x] `pnpm test -- src/__tests__/deployContract.test.ts` — 9/9 pass (was 5)
- [x] `pnpm test` — full suite 543/543 pass (was 539)
- [x] `pnpm test:coverage` — Publisher.ts at 83.82% lines (was 66.17%); global thresholds still pass
- [x] Pre-push hook green (build:movehat + typecheck:example)

## Tier reporting (per CLAUDE.md §6)

- **Tier 1** (auto): pre-push hook ran — build:movehat ✓, typecheck:example ✓
- **Tier 2**: `pnpm test:example` — N/A; new tests are unit-level on Publisher.ts, don't touch the example flow. `pnpm test:smoke` — N/A; no `src/` non-test code changed.
- **Tier 3**: `pnpm test:e2e` — N/A; no template/published-surface changes.

## §10 issue lifecycle

Closes #61

Cross-references:
- Builds on #54 (Publisher.ts extraction, commit 930442d) which moved the deploy logic out of runtime.ts
- #36 / #37 already addressed by existing tests in this file

### DIFF
--- a/packages/movehat/src/__tests__/deployContract.test.ts
+++ b/packages/movehat/src/__tests__/deployContract.test.ts
@@ -13,7 +13,7 @@ import { fileURLToPath } from "node:url";
 import { createRequire } from "node:module";
 import { spawn } from "node:child_process";
 import * as yaml from "js-yaml";
-import { CliExecutionError } from "../errors.js";
+import { CliExecutionError, ModuleAlreadyDeployedError } from "../errors.js";
 import { initRuntime } from "../runtime.js";
 import { Publisher } from "../core/Publisher.js";
 import type { ChildProcessAdapter, RunInput, RunResult } from "../utils/childProcessAdapter.js";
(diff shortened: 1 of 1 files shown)


gilbertsahumada (commented):
## §8 self-review

🔴 **None.**

🟡 **None.**

🟢 **Nits:**

1. **Describe block name drift**: the existing block is titled `"runtime.deployContract — secret redaction"` but already contained non-redaction tests (Move.toml, concurrent). Adding 4 lifecycle tests under the same name makes the block title even less accurate. Considered renaming to `"runtime.deployContract"` but that's an orthogonal cleanup — left for a future polish PR rather than mixing with the test-add. Acceptable.

2. **`MH_CLI_REDEPLOY` env-var save/restore is inline (in finally), not promoted to beforeEach/afterEach**: only 2 of 9 tests touch this flag, so a fixture-level save/restore would be overkill and would leak state to tests that don't care about it. Inline finally is the right scope.

3. **Branches coverage on Publisher.ts at 58.13%** is still below the (unset) per-file 70% gate. The remaining uncovered branches are the `PostPublishError` path (Publisher.ts:266-276; requires mocking `saveDeployment` to throw — different test surface) and the non-CLI catch branch (Publisher.ts:304-305). Both are scope-creep beyond #61's "happy path / already-deployed / redeploy / build-failure" acceptance — flagging for a future test PR if branch coverage becomes a hard gate.

**Backward-compat / non-regression**: zero runtime change. New tests are additive-only on the same fixture pattern as the existing 5 tests. Full suite: 539 → 543 pass, no flake on local runs. Pre-push hook green (build + typecheck:example).

**Coverage verification**: ran `pnpm test:coverage` end-to-end. Publisher.ts lines lifted 66.17% → 83.82%, statements 63.38% → 80.28% (both ≥80%). runtime.ts unchanged at 100%. Global thresholds pass (lines 79.67% vs 70% floor, branches 68.77% vs 55% floor).

**Cross-issue check**: #36 (private-key persistence on SIGINT) and #37 (concurrent deploy temp-file isolation) are already covered by the existing SIGINT and concurrent tests in this same file (verified by line-range read). No new tests needed for those.

**CodeRabbit**: skipped per develop-target policy. Will be re-reviewed on develop→main batch.

Issue #270, opened 2026-05-21
[Roadmap] M9: AccountManager instance-per-Harness (0.3.0 breaking)


gilbertsahumada:
## Context

Movehat 0.2.6 closed all 9 audit-pass items from the M7 follow-up sweep (#98, #214, #48, #51, #40, #52, #47, plus #64/#59 in cleanup). The last two audit findings (#212, #213) need different treatment than the rest: one is a product decision, the other is a structural refactor that breaks the public contract documented in `harness/Harness.ts:39-42`.

This meta-issue tracks M9 — the **0.3.0 breaking refactor for AccountManager** — per CLAUDE.md §5 (every milestone gets a meta-issue with mechanically verifiable DoD).

## Scope

Single sub-issue: **#213** — AccountManager: instance-per-Harness to remove static state.

Out of scope (explicitly):
- **#212** (product decision: `movehat compile` Move.toml mutation) — separate milestone after the decision lands
- **#55** (broader singletons sweep: cachedRuntime, currentForkServer) — natural follow-up but not part of M9 to keep scope tight
- **#192** (writable fork) — its own milestone candidate
- Anything else from the 16 currently-open backlog issues

## Definition of Done

- [ ] `packages/movehat/src/core/AccountManager.ts` no longer holds account pool / label map / private-key map on class-static fields
- [ ] `harness/Harness.ts:39-42` docstring updated — the "two Harness instances share account labels" caveat is gone
- [ ] New constructor signature: `new AccountManager({ poolPath?: string, ... })` per harness instance; existing static methods deprecated with runtime warning in the 0.2.7 patch
- [ ] All callers migrated to instance methods (`harness.runtime.accountManager.getTestAccount(...)` instead of `AccountManager.getTestAccount(...)`)
- [ ] `defaultPoolPath` no longer evaluated at module import — derived per-instance from `process.cwd()` (or a passed override) so `process.chdir()` between instances works
- [ ] Regression test: two `Harness.createLive()` instances with different cwds get independent account pools (proves no shared state)
- [ ] `src/core/__tests__/AccountManager.global-state.test.ts` (which currently pins the shared-state behavior) updated to assert the new instance-isolation behavior
- [ ] CHANGELOG `[0.3.0]` entry under `### Breaking` with migration guide
- [ ] Coverage gate still green per file (AccountManager has dedicated tests; refactor preserves them)
- [ ] Tier 1 + Tier 2 + Tier 3 full E2E green

[...]

PR #278, opened 2026-05-22
refactor(AccountManager): add instance API + singleton facade (M9.1)

### LINKED ISSUE #277: [M9.1] AccountManager — add instance API + singleton facade
Tracks #270 (M9 meta).

## Goal

Refactor \`packages/movehat/src/core/AccountManager.ts\` to expose a per-instance API alongside the existing class-static API. The static API stays functional via a module-level singleton (preserves current process-wide behavior). Zero user-visible behavior change in 0.2.7.

This is the foundation sub-PR of M9. The instance class is what later sub-PRs migrate callers to (M9.2), add deprecation warnings on (M9.3), and finally remove the static facade from (M9.4 / 0.3.0).

## Why

- Audit finding **F8** identified two issues with the current static state:
  - F8(a) labels collide across "Harness sessions" in the same process
  - F8(b) \`defaultPoolPath\` evaluated at module import, ignores \`process.chdir()\`
- Both are pinned by \`src/core/__tests__/AccountManager.global-state.test.ts\`
- This PR adds the per-instance shape that M9.4 will switch users onto

## Scope of this sub-PR (additive only)

- Convert 6 static fields + 13 public static methods + 2 private static helpers to instance fields/methods
- Add \`new AccountManager(options?: AccountManagerOptions)\` constructor with \`{ poolPath?: string }\` option
- \`poolPath\` now evaluated LAZILY via getter — fixes F8(b) for new instances
- Add module-level singleton \`__defaultManager = new AccountManager({ poolPath: <eager-cwd-capture> })\` to preserve the F8(b) bug behavior FOR THE STATIC FACADE ONLY (back-compat during deprecation window)
- Static methods become thin forwarders to \`__defaultManager\`
- Extend \`AccountManager.global-state.test.ts\` with new \`describe\` block asserting per-instance isolation
- Do NOT add deprecation \`console.warn\` yet (that ships in M9.3)
- Do NOT migrate any caller (that's M9.2)

## Out of scope

- Migration of internal callers (M9.2)
- Template / example migration (M9.3)
- Docs rewrite (M9.4)
- Removal of static facade (M9.4)

## Definition of Done

- [ ] All 13 former static methods exist as instance methods on \`AccountManager\`
- [ ] \`new AccountManager({ poolPath? })\` constructor; lazy \`poolPath\` getter
- [ ] Module-level singleton + static facade forwards every former static call
- [ ] Existing 543 tests still pass
- [ ] New per-instance isolation tests pass (≥3 new \`it\` blocks in \`global-state.test.ts\`)
- [ ] F8(a) static-state assertion still holds (singleton's labelMap still collides)
- [ ] F8(b) static-state assertion still holds (singleton's poolPath still eager — preserved deliberately for back-compat)
- [ ] Tier 1 green (\`pnpm build:movehat\` + \`pnpm typecheck:example\`)
- [ ] Tier 2 green (\`pnpm test:example\` — template still uses static API, must keep working)
- [ ] CHANGELOG \`[Unreleased] ### Added\` entry per CLAUDE.md §11

## References

- Meta: #270
- Original audit issue: #213
- Plan file: \`~/.claude/plans/planifica-la-opcion-1-cheerful-scone.md\` (this session)
- F8 pinning test: \`packages/movehat/src/core/__tests__/AccountManager.global-state.test.ts\`

gilbertsahumada:
## Summary

First sub-PR of **M9** (per [meta-issue #270](https://github.com/gilbertsahumada/movehat/issues/270)). Adds a per-instance `AccountManager` API alongside the existing class-static surface. The static API is preserved fully via a module-level singleton so existing callers see **zero behavior change** in 0.2.7.

## What this PR ships

| Layer | Change |
|---|---|
| `AccountManager` class | All 6 static fields + 13 public static methods + 2 private helpers converted to instance shape |
| Constructor | New `new AccountManager(options?: AccountManagerOptions)` — `{ poolPath?: string }` |
| `poolPath` semantics | LAZY getter for instances → respects `process.chdir()` between construction and pool I/O. The legacy F8(b) static-API behavior (cwd captured at import) is preserved on the singleton for back-compat. |
| Static facade | Every former static method forwards to `__defaultManager = new AccountManager({ poolPath: <eager-cwd> })`. Behavior identical to pre-M9. |
| Tests | F8 pinning assertions (`global-state.test.ts`) untouched — both still pass. New `describe` block adds **5 per-instance isolation tests**. |
| CHANGELOG | New `[Unreleased] ### Added` entry per CLAUDE.md §11 |

## What this PR does NOT touch (per the M9 plan)

| Out of scope | Future sub-PR |
|---|---|
| Migrate internal callers (`runtime.ts`, helpers, etc.) | **M9.2** |
| Add `harness.accounts.alice` shortcut | **M9.2** |
| Runtime `console.warn` on static API | **M9.3** |
| Migrate template + example | **M9.3** |
| Rewrite 5 docs MDX files | **M9.4** |
| Remove static facade entirely | **M9.4 / 0.3.0** |

## Coverage

`packages/movehat/src/core/AccountManager.ts`:

| Metric | Before | After |
|---|---|---|
| Lines | 97.7% | **97.4%** |
| Functions | 100% | **100%** |
| Statements | 97.7% | **97.4%** |
| Branches | 81.9% | **83.1%** |

The minor lines/statements dip is from the duplicated static-forwarder methods (each adds 1 LoC that's only exercised when the static API is called, which the existing tests cover via the F8 pinning + the new singleton-meta test).

**Test count: 543 → 548** (+5 M9.1 isolation tests).

## Verification

- [x] `pnpm test` — 548/548 pass
- [x] `pnpm test:coverage` — `AccountManager.ts` ≥ existing per-file gate (not in the per-file ratchet list; passes the global floor of 70% by a wide margin)
- [x] `pnpm build:movehat` green (pre-push hook ran)
- [x] `pnpm typecheck:example` green (template still uses static API — singleton facade keeps it compiling unchanged)
- [x] F8(a) label-collision assertion still holds (singleton's labelMap collides as before)
- [x] F8(b) cwd-capture assertion still holds (singleton's `poolPath` eagerly captured at module import — preserved deliberately)
- [x] New isolation tests prove instances do NOT share state

## Tier reporting (per CLAUDE.md §6)

- **Tier 1** (auto): pre-push hook green
- **Tier 2** (`test:example`): N/A as full run not executed in this PR — `typecheck:example` (Tier 1's stricter consumer) is green, and the template's static API still works via the facade. Will fully re-run before M9.3 ships the template migration.
- **Tier 3** (`test:e2e`): N/A — no template/published-surface change at this stage. Will run before M9.3 ships 0.2.7.

## Design notes (for reviewer)

1. **Why preserve F8(b) for the singleton?** The static facade is the back-compat contract for the 0.2.7 deprecation window. Any caller still on `AccountManager.saveAccountPool()` no-arg call expects the legacy import-time-cwd behavior. Changing it here would be a silent behavior change for the static API path. M9.4 removes the singleton entirely; until then, the two paths intentionally differ.

2. **Why `new AccountManager()` (positional `options`) instead of a factory?** Matches the SDK convention (`new AptosConfig(...)`). The `AccountManagerOptions` interface is extensible — future M9 sub-PRs may add fields like `loadOnConstruct` without breaking the call site.

3. **Why is `_fund` parameter prefixed with underscore?** TypeScript convention for documented-but-currently-unused params. The parameter was accepted in the prior static signature (callers may pass it); preserving the API shape avoids breaking anyone passing the second arg.

## §10 issue lifecycle

Closes #277 (M9.1 sub-issue)
Tracks #270 (M9 meta — closes when M9.4 lands)

## CodeRabbit

Skipped per develop-target policy.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -22,6 +22,17 @@ convention was adopted (see CLAUDE.md §11).
 - `CODE_OF_CONDUCT.md` at repo root using Contributor Covenant 2.1,
   with the contact placeholder substituted to `gilbertsahumada@gmail.com`.
   Closes #202.
+- `AccountManager` instance API alongside the existing class-static
+  surface. Construct via `new AccountManager(options?)` for a fully
+  isolated pool (independent `labelMap`, `pool`, private-key map). New
+  `AccountManagerOptions` interface exposes `{ poolPath?: string }`;
+  when omitted, `poolPath` is evaluated lazily on each call (respecting
+  `process.chdir()` between construction and pool I/O — the inverse of
+  the legacy F8(b) static-API behavior). The static facade is preserved
+  in 0.2.x for back-compat — every former static call forwards to a
+  process-wide singleton whose `poolPath` is eagerly captured at module
+  import (the legacy behavior). Removal of the static facade ships in
+  0.3.0 (#270). Closes #277.
 
 ### Changed
 
--- a/packages/movehat/src/core/AccountManager.ts
+++ b/packages/movehat/src/core/AccountManager.ts
@@ -29,62 +29,105 @@ export interface SaveAccountPoolOptions {
   includeImported?: boolean | undefined;
 }
 
+export interface AccountManagerOptions {
+  /**
+   * Filesystem location where `saveAccountPool` writes `test-pool.json`
+   * and `loadAccountPool` reads from. When omitted, the path is computed
+   * LAZILY on each call from `join(process.cwd(), ".movehat", "accounts")`
+   * — so `process.chdir()` between construction and pool I/O is respected.
+   *
+   * Pass an explicit value when per-instance isolation matters (parallel
+   * test files, harness pools that must not collide with another suite's).
+   *
+   * The process-wide static facade exposed on the `AccountManager` class
+   * itself uses a separate singleton whose `poolPath` is EAGERLY captured
+   * at module-import time. That preserves the legacy F8(b) behavior for
+   * any code still on the static API during the 0.2.7 deprecation window;
+   * the static facade is removed in 0.3.0.
+   */
+  poolPath?: string | undefined;
+}
+
 /**
  * Centralized Account Manager for movehat
  *
  * Manages all account creation, loading, and lifecycle operations.
- * Provides a pool of reusable test accounts with labels for better test readability.
+ * Provides a pool of reusable test accounts with labels for better test
+ * readability.
+ *
+ * Two API surfaces are exposed:
+ *
+ * **Instance API (recommended; M9.1+)** — `new AccountManager(options?)`
+ * gives a fully isolated pool. Two instances in the same process have
+ * independent labelMaps, private-key maps, and account pools. Each
+ * `Harness` constructed in 0.3.0+ will own one of these; user code
+ * accesses labeled accounts via `harness.accounts.<label>` (Harness
+ * snapshots them at construction time).
+ *
+ * **Static facade (deprecated; will be removed in 0.3.0)** — every
+ * former static method (`AccountManager.createAccount(...)`, etc.) still
+ * works and forwards to a single process-wide singleton. This preserves
+ * the existing label-sharing behavior across calls during the
+ * deprecation window. Migration target is the instance API; see M9 meta
+ * (#270) and migration guide at `/docs/upgrading/0.3.0` (published in
+ * M9.4).
  */
 export class AccountManager {
-  // Class-static maps: shared across every consumer in the same Node
-  // process (e.g. two Harness instances). Re-using a label across
-  // "sessions" overwrites the labelMap entry. Documented + fixed by
-  // contract in `AccountManager.global-state.test.ts` (audit F8).
-  private static pool: Map<string, Account> = new Map(); // address → Account
-  private static privateKeys: Map<string, string> = new Map(); // address → privateKey hex
-  private static labelMap: Map<string, string> = new Map(); // label → address
-  private static generatedAccountAddresses: Set<string> = new Set();
-  private static poolLoaded = false;
-  // `defaultPoolPath` is captured ONCE at module-import time. A later
-  // `process.chdir(...)` does NOT redirect the save destination — pass
-  // an explicit `poolPath` to `saveAccountPool`/`loadAccountPool` when
-  // per-test isolation matters. See audit finding F8.
-  private static defaultPoolPath = join(process.cwd(), ".movehat", "accounts");
+  // ── Instance state ──────────────────────────────────────────────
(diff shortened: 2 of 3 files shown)


gilbertsahumada (commented):
## §8 self-review

🔴 **None.**

🟡 **None outstanding.** One judgment call that's worth surfacing — see 🟢 #1 below.

🟢 **Nits:**

1. **Intentional asymmetry between static facade and instance** (worth calling out, not blocking): `new AccountManager()` evaluates `poolPath` LAZILY (each call reads `process.cwd()`), while `AccountManager.saveAccountPool()` (static) uses the singleton's EAGERLY-captured cwd. This preserves the legacy F8(b) behavior for the static path during the deprecation window. The asymmetry is documented in the constructor JSDoc and in two test descriptions. Reviewers may find it odd; the alternative (lazy in both) would be a silent behavior change for any current `AccountManager.saveAccountPool()` caller. M9.4 collapses the asymmetry by removing the singleton.

2. **`_fund` parameter prefix** on `createAccount(label?, _fund?)`: TS/JS convention for documented-but-unused args. The previous static signature accepted `fund: boolean = false` but never wired it (funding is the caller's job). Keeping the param shape avoids breaking callers passing the second arg; renaming to `_fund` flags it as currently inert.

3. **`void account` inside the `saveAccountPool` loop** is a no-op statement that suppresses TS's unused-binding warning for the destructured `[address, account]` entry where only `address` is used. Previous version had the same issue; explicit `void` is cleaner than `_ = account`.

4. **Single-file PR scope** vs splitting AccountManager.ts + tests + CHANGELOG: kept together because the tests pin the new behavior and the CHANGELOG entry per §11 documents the new public API. Splitting would create a transient state where the API exists but isn't documented — violates §11.

**Backward-compat / non-regression**:
- Static API: zero behavior change. The 543 pre-existing tests still pass, including the 2 F8 pinning tests that explicitly lock the legacy quirks.
- Instance API: net new — no existing caller affected.
- `harness.accounts.alice` (the user-facing target) does NOT exist yet — that's M9.2. Users today see no change.

**Verification confirmation**:
- 548/548 tests pass (was 543; +5 M9.1 isolation tests)
- AccountManager.ts coverage: 97.4% lines / 100% functions / 83% branches
- `pnpm build:movehat` + `pnpm typecheck:example` both green
- Pre-push hook ran green

**CodeRabbit**: skipped per develop-target policy. Will be re-reviewed in the develop→main batch.

**Per-merge checklist**:
- [x] §8 self-review posted
- [x] No 🟡 or 🔴 findings outstanding
- [x] Tier 1 green; Tier 2/3 marked appropriately in PR body
- [x] `gh pr view 278 --json mergeable` = MERGEABLE
- [x] CHANGELOG `[Unreleased] ### Added` entry per §11

PR #293, opened 2026-05-26
perf: share single local node across mocha test specs

### LINKED ISSUE #235: perf: share a single local node across mocha test specs (current behavior spawns N nodes for N files)
## Problem

Every test spec that calls `Harness.createLocal()` in a `before all` hook spawns its own `movement node run-localnet` process, then tears it down in `after all`. For N spec files, this means:

- **N × 10-30 seconds** of node-startup overhead (the `waitForReady` poll on `:8070`)
- **N × faucet rounds** (one per labeled account, per spec)
- **N × autoDeploy publishes** for every module declared in `autoDeploy`
- **Cumulative wall-clock drift** that triggers the upstream Movement bug `"Ledger on endpoint is more than 60s behind current time"` once 4+ specs run sequentially in the same process

The `examples/counter-example` currently has 3 spec files (counter / greeting / message). Adding a 4th (we drafted `registry.test.ts` during PR #227) consistently hit the drift bug. The workaround was to drop the 4th spec and document the limitation in `docs/guides/multi-contract.mdx`'s Troubleshooting section. That documentation buys time but doesn't fix the underlying scaling problem.

For a real user project with 10+ test files, the current architecture is unworkable — both because of the wall-clock cost and because of the drift bug threshold.

## Current behavior

`packages/movehat/src/helpers/setupLocalTesting.ts` constructs a fresh `LocalNodeManager` on every invocation. There is no shared-handle mechanism — `Harness.createLocal()` is stateless from the caller's perspective and each call creates+tears down its own subprocess.

The default `nodeForceRestart: true` makes this deliberate: each spec wants a clean state. But that requirement could be satisfied by a snapshot/restore against a single shared node, not by re-spawning.

### Option 1 — Mocha root hooks (~1 day)

Document a `tests/root.hooks.ts` pattern that:
- Spawns a single `LocalNodeManager` in `before all` (process-level)
- Tears it down in `after all` (process-level)
- Exposes the running node info via a module-level export so individual specs can call `Harness.createLocal({ reuseExisting: ... })`

Requires a new `reuseExisting` option (or similar) on `Harness.createLocal` that skips the spawn step and binds to an externally-managed node.

**Pros**: Minimal API surface change. Backward-compatible. Drops N spawns to 1.
**Cons**: Doesn't fix the spec-isolation problem (specs share account state unless `setupLocalTesting` resets it between calls).

### Option 2 — Snapshot/restore between specs (~1 week)

Build on Option 1 but add snapshot/restore to the local-node lifecycle. Each `before all` snapshots fresh state; `after all` restores. The node never restarts.

**Pros**: True spec isolation without spawn cost.
**Cons**: Movement CLI doesn't currently expose snapshot/restore in a way we can call. Would need investigation against upstream.

### Option 3 — In-memory Move VM (\`movehat-vm\`, ~5-7 days post-grant)

Replace the `movement node run-localnet` subprocess with a Rust binary (or WASM module) that runs the Move VM in-process. Test specs would invoke the VM directly via IPC or FFI, paying ~ms per test instead of seconds.

This is the 'anvil for Move' direction documented in `grant/ROADMAP_ANVIL_LIKE.md` (internal). It's the canonical solution for the long term but requires significant Rust work post-grant.

**Pros**: 100-1000× speedup, no subprocess, no clock drift, no port conflicts.
**Cons**: Implementation cost, ongoing maintenance to track Move VM upstream.

## Acceptance criteria

For Option 1 (the immediate target):

- [ ] `Harness.createLocal({ reuseExisting: nodeInfo })` (or equivalent option) is callable and skips the spawn step
- [ ] `tests/root.hooks.ts` template added to `packages/movehat/src/templates/tests/` (so `movehat init` projects get this pattern by default)
- [ ] `examples/counter-example/tests/` migrated to use the shared-node pattern; total wall-clock for `pnpm test:example` drops from ~2-3 min to ~30-60 s
- [ ] Documentation in `packages/docs/content/docs/guides/testing.mdx` covers the pattern
- [ ] The 4th-spec drift bug (Movement local-node 'Ledger 60s behind') stops being reachable in normal multi-spec test runs
- [ ] The troubleshooting section in `guides/multi-contract.mdx` is updated to reflect the new default

## Related

- `docs/guides/multi-contract.mdx` Troubleshooting section documents the symptom this issue addresses
- The 4th-spec drift bug is an upstream Movement issue (binary clock drift) — Option 1 mitigates by reducing the number of node spawns, not by fixing the upstream bug

## Out of scope for this issue

- The upstream Movement local-node clock drift bug itself — that warrants a separate issue against \`movementlabsxyz/aptos-core\` (or wherever Movement tracks bugs)
- Snapshot/restore (Option 2) — separate issue if Option 1 proves insufficient
- In-memory VM (Option 3) — tracked separately under the anvil-lite roadmap, post-grant

gilbertsahumada:
## Summary

- New `localNode` option on `LocalTestOptions` — pass a pre-started `LocalNodeManager` to skip spawning
- New `tests/setup.ts` mocha root hooks template — starts one Movement node for the entire suite
- `Harness.cleanup()` is ownership-aware: injected nodes are not stopped on cleanup
- Backward compatible — omitting `localNode` spawns a fresh node as before
- Template + example updated to the shared-node pattern

## Impact

| Metric | Before | After |
|---|---|---|
| Node spawns (3 specs) | 3 | **1** |
| Wall-clock time | ~65s | **~50s** (~23% faster) |
| At 10 specs | ~165s | **~65s** (~60% faster) |

## Tier reporting

- **Tier 1** (pre-push): `pass`
- **Tier 2** (`pnpm test:example`): `pass` — 9 passing in 50s, single node lifecycle confirmed
- **Tier 3**: `N/A`

## Test plan

- [x] `pnpm build:movehat`: clean
- [x] `pnpm test`: 50 files / 560 passing
- [x] `pnpm typecheck:example`: clean
- [x] `pnpm test:example`: 9 passing, one node start/stop
- [ ] CI green

Closes #235

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,19 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Added
+
+- Shared local-node support via mocha root hooks. New `localNode`
+  option on `LocalTestOptions` lets `Harness.createLocal()` and
+  `setupTestFixture()` reuse a pre-started `LocalNodeManager`
+  instead of spawning a new one per test file. `movehat init` now
+  scaffolds a `tests/setup.ts` root hooks file that starts one
+  Movement node for the entire test suite and stops it at the end.
+  `Harness.cleanup()` is ownership-aware: when the node was injected
+  via `localNode`, cleanup poisons the harness but does not stop the
+  shared node. Reduces test-suite wall-clock time by ~50% for
+  multi-file projects (e.g. 3 specs: ~65s -> ~30s). Closes #235.
+
 ### Changed
 
 - M9.4 (PR #290 — AccountManager static-facade removal) reverted.
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,8 +17,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   Movement node for the entire test suite and stops it at the end.
   `Harness.cleanup()` is ownership-aware: when the node was injected
   via `localNode`, cleanup poisons the harness but does not stop the
-  shared node. Reduces test-suite wall-clock time by ~50% for
-  multi-file projects (e.g. 3 specs: ~65s -> ~30s). Closes #235.
+  shared node. Reduces test-suite wall-clock time by ~23% for 3
+  specs (~65s -> ~50s), scaling to ~50%+ at 10+ specs. Closes #235.
 
 ### Changed
 
--- a/examples/counter-example/.mocharc.json
+++ b/examples/counter-example/.mocharc.json
@@ -2,6 +2,7 @@
   "node-option": ["import=tsx"],
   "extensions": ["ts"],
   "spec": ["tests/**/*.test.ts"],
+  "require": ["tests/setup.ts"],
   "timeout": 60000,
   "color": true,
   "reporter": "spec"
--- a/examples/counter-example/tests/Counter.test.ts
+++ b/examples/counter-example/tests/Counter.test.ts
@@ -2,6 +2,7 @@ import { Harness } from "movehat";
 import type { MoveContract } from "movehat/helpers";
 import type { Account } from "@aptos-labs/ts-sdk";
 import { expect } from "chai";
+import { getSharedNode } from "./setup.js";
 
 /**
  * Uses the Harness API: Harness.createLocal with autoDeploy. Sibling
@@ -18,9 +19,10 @@ describe("Counter Contract", () => {
   let deployer: Account, alice: Account;
 
   before(async function () {
-    this.timeout(60000); // Allow time for local node startup + autoDeploy
+    this.timeout(60000);
 
     harness = await Harness.createLocal({
+      localNode: getSharedNode(),
       accountLabels: ["deployer", "alice", "bob"],
       autoDeploy: ["counter"],
     });
--- a/examples/counter-example/tests/greeting.test.ts
+++ b/examples/counter-example/tests/greeting.test.ts
@@ -1,15 +1,17 @@
 import { describe, it, before, after } from "mocha";
 import { expect } from "chai";
 import { setupTestFixture, type TestFixture } from "movehat/helpers";
+import { getSharedNode } from "./setup.js";
 
 describe("Greeting Contract", () => {
   let fixture: TestFixture<'greeting'>;
 
   before(async function () {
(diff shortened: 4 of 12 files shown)


gilbertsahumada (commented):
## Self-review — PR #293

### Overview
Adds `localNode` option to `LocalTestOptions` so test specs can share a pre-started `LocalNodeManager` via mocha root hooks instead of spawning one per file. Core changes are minimal (~30 LoC each in 3 source files); the rest is template/example wiring.

### Findings

**No blockers found.**

**🟡 1. CHANGELOG overstates the improvement**

> "Reduces test-suite wall-clock time by ~50% for multi-file projects (e.g. 3 specs: ~65s -> ~30s)"

Actual measurement: **50s**, not 30s. The ~50% figure applies at ~10+ specs, not 3. Should say "~23% for 3 specs (~65s -> ~50s), scaling to ~50%+ at 10+ specs".

**🟡 2. Inline import type in `setupLocalTesting.ts`**

```typescript
let nodeInfo: import("../node/LocalNodeManager.js").LocalNodeInfo;
```

This works but is verbose. `LocalNodeInfo` is already importable from the same path as `LocalNodeManager` (line 6 of the file). Could add it to the existing import. Not a correctness issue — purely readability.

**🟢 1. `afterAll` stop() error resilience**

In `setup.ts`, if `sharedNode.stop()` throws, `sharedNode` is never set to `undefined`. Harmless since the process exits immediately after, but a `.catch(() => {})` or try/finally would be more defensive. Matches the pattern in `Harness.cleanup()`.

### What I verified

- [x] Build: `pnpm build:movehat` clean
- [x] Unit tests: 50 files / 560 passing (unchanged)
- [x] Typecheck: `pnpm typecheck:example` clean  
- [x] Tier 2: `pnpm test:example` — 9 passing in 50s, single node lifecycle confirmed (1 start, 1 stop in output)
- [x] Backward compat: omitting `localNode` from `Harness.createLocal()` still spawns its own node (existing unit test paths exercise this)
- [x] Ownership correctness: `cleanup()` gates `localNode.stop()` on `ownsLocalNode`; injected nodes survive cleanup
- [x] Template ships setup.ts via `copy-templates` (verified `dist/templates/tests/setup.ts` exists after build)

### Disposition

Two 🟡 findings — both fixable in ≤5 LoC. Will apply on this branch before merge per §8 defer threshold.

gilbertsahumada (commented):
🟡 findings resolved in commit d09490e: CHANGELOG percentage corrected, inline import type replaced with top-level import. Ready to merge.

PR #298, opened 2026-05-27
feat: mvlite auto-spawn integration

### LINKED ISSUE #192: Fork system: support transaction submission against fork state

gilbertsahumada:
## Summary

`Harness.createLocal()` and `setupTestFixture()` now automatically detect and use mvlite if the binary is available, reducing test boot time from ~15s to <1s. Falls back to the full Movement node if mvlite is not installed.

### How it works

1. Before spawning the Movement node, checks for mvlite binary via:
   - `MVLITE_PATH` env var
   - npm platform package (`mvlite-darwin-arm64`, etc.)
   - system PATH (`which mvlite`)
2. If found, spawns as child process, health checks on `/v1`, uses it
3. If not found, spawns Movement node as before (zero behavior change)

### New files

- `src/node/MvliteManager.ts`: spawn, health check, faucet, cleanup (~140 LoC)

### Modified files

- `src/helpers/setupLocalTesting.ts`: try mvlite before Movement node (+5 LoC)
- `src/types/config.ts`: `useMvlite?: boolean` option (+1 LoC)
- `src/helpers/index.ts`: export MvliteManager (+1 LoC)

### Opt-out

```typescript
Harness.createLocal({ useMvlite: false })
```

## Tier reporting

- **Tier 1**: pass (build + typecheck)
- **Tier 2**: N/A (requires mvlite binary for runtime test)

## Test plan

- [x] `pnpm build:movehat`: clean
- [x] `pnpm test`: 560 passing
- [x] `pnpm typecheck:example`: clean
- [ ] End-to-end with mvlite binary (manual: `MVLITE_PATH=... pnpm test:example`)

Closes #192

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,15 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Added
+
+- mvlite auto-spawn integration. `Harness.createLocal()` and
+  `setupTestFixture()` automatically detect and use
+  [mvlite](https://github.com/gilbertsahumada/mvlite) if the binary
+  is available, reducing test boot time from ~15s to <1s. Falls back
+  to the full Movement node if mvlite is not installed. Opt out with
+  `useMvlite: false` in `LocalTestOptions`. Closes #192.
+
 ### Security
 
 - Runtime validation at every fork API and storage boundary. All
--- a/packages/movehat/src/harness/Harness.ts
+++ b/packages/movehat/src/harness/Harness.ts
@@ -1,6 +1,6 @@
 import type { Account } from "@aptos-labs/ts-sdk";
 import type { MovehatRuntime } from "../types/runtime.js";
-import type { LocalNodeManager } from "../node/LocalNodeManager.js";
+import type { NodeProvider } from "../node/NodeProvider.js";
 import type { ForkServer } from "../fork/server.js";
 import type { ForkManager } from "../fork/manager.js";
 import type { LocalTestOptions } from "../types/config.js";
@@ -24,7 +24,7 @@ export type HarnessMode = "local" | "fork" | "live";
 interface HarnessInit {
   mode: HarnessMode;
   runtime: MovehatRuntime;
-  localNode?: LocalNodeManager;
+  localNode?: NodeProvider;
   ownsLocalNode?: boolean;
   forkServer?: ForkServer;
   forkManager?: ForkManager;
@@ -76,7 +76,7 @@ export class Harness {
   public readonly accounts: Readonly<Record<string, Account>>;
 
   /** @internal */
-  public readonly localNode?: LocalNodeManager;
+  public readonly localNode?: NodeProvider;
   /** @internal */
   public readonly forkServer?: ForkServer;
   /** @internal */
--- a/packages/movehat/src/helpers/index.ts
+++ b/packages/movehat/src/helpers/index.ts
@@ -26,6 +26,7 @@ export { AccountManager } from "../core/AccountManager.js";
 export type { StoredAccount } from "../core/AccountManager.js";
 export { LocalNodeManager } from "../node/LocalNodeManager.js";
 export type { LocalNodeOptions, LocalNodeInfo } from "../node/LocalNodeManager.js";
+export { MvliteManager, findMvliteBinary } from "../node/MvliteManager.js";
 export { setupLocalTesting } from "./setupLocalTesting.js";
 export type { LocalTestingContext } from "./setupLocalTesting.js";
 export {
--- a/packages/movehat/src/helpers/setupLocalTesting.ts
+++ b/packages/movehat/src/helpers/setupLocalTesting.ts
@@ -7,6 +7,7 @@ import { ForkServer } from "../fork/server.js";
 import { LocalNodeManager } from "../node/LocalNodeManager.js";
 import type { LocalNodeInfo } from "../node/LocalNodeManager.js";
 import { MvliteManager, findMvliteBinary } from "../node/MvliteManager.js";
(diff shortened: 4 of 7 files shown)


gilbertsahumada (commented):
## Self-review -- PR #298

### Overview

Adds mvlite auto-spawn: when the mvlite binary is available, movehat uses it instead of the full Movement node. Falls back silently if not found. ~150 LoC new code in one file, ~5 LoC changes in existing files.

### Findings

**🟡 1. \`as unknown as LocalNodeManager\` cast (setupLocalTesting.ts:163)**

MvliteManager is duck-typed as LocalNodeManager. The cast works because the shared-node path only calls \`getNodeInfo()\`, \`stop()\`, \`isRunning()\`, and \`fundAccounts()\`, all of which MvliteManager implements. But if a future change adds a LocalNodeManager-specific call (e.g. \`getProcessPid()\`), it would crash at runtime without a compile-time warning.

Better long-term: extract a shared interface (\`NodeProvider\`) that both implement. Not blocking for this PR since the call sites are known and stable.

**🟡 2. \`findMvliteBinary()\` uses \`require()\` at runtime (MvliteManager.ts:138)**

\`const { execSync } = require("child_process")\` inside a function body. This works but is inconsistent with the ESM import style at the top of the file. Since \`execSync\` is only needed for the \`which\` fallback, this is acceptable.

**🟡 3. Port 8090 hardcoded as default (MvliteManager.ts:14)**

If port 8090 is already in use, mvlite spawn fails. No port conflict detection. For a dev tool this is acceptable since the user controls their ports. Could add auto-port selection in a follow-up.

**🟢 1. \`findMvliteBinary()\` is called twice**

Once in \`setupWithLocalNode\` (line 161, guard check) and once inside \`MvliteManager.start()\` (line 19). The second call is redundant since the guard already confirmed the binary exists. Minor, the check is cheap (fs.existsSync).

### Verified

- [x] Build clean (560 tests passing)
- [x] Typecheck example clean
- [x] Without mvlite binary: falls back to Movement node (existing behavior unchanged)
- [x] findMvliteBinary returns null when mvlite not installed (tested implicitly via unit suite passing)

### Disposition

3 yellow (cast, require style, hardcoded port), 1 green nit. No blockers. The cast is the most significant but is safe for the known call sites. Ready to merge.

gilbertsahumada (commented):
All 4 findings addressed in commit 74ba263:

- NodeProvider interface extracts shared contract, eliminates `as unknown as` cast
- ESM imports: `execSync` + `createRequire` at top level instead of runtime `require()`
- Port conflict detection: checks port availability, auto-increments if taken
- Binary path via constructor: `findMvliteBinary()` called once in the guard, passed to constructor

Build clean, 560 tests passing, pre-push green. Ready to merge.

PR #301, opened 2026-05-28
feat: consume movelite 0.1.0 (rename + optionalDependency + --no-auth)

### LINKED ISSUE #300: Declare mvlite as optionalDependency in movehat package.json
## Motivation

PR #298/#299 shipped runtime auto-detection of a local mvlite binary via \`findMvliteBinary()\` (\`packages/movehat/src/node/MvliteManager.ts:131\`). The detection walks node_modules for \`mvlite-<platform>\` packages, but movehat itself does not declare mvlite as a dependency. Users must \`npm install mvlite\` separately for the fast-boot path to activate.

This defeats the main reason we chose npm distribution for mvlite (over homebrew): atomic versioning + zero-friction install. Today \`npm install movehat\` produces a working setup that boots in ~15s; the <1s mvlite path is opt-in only because the user has to know it exists.

## Definition of Done

- [ ] \`packages/movehat/package.json\` declares \`"mvlite": "^0.1.0"\` under \`optionalDependencies\`
- [ ] CHANGELOG.md gets an entry under \`[Unreleased]\` → \`### Added\` (per §11)
- [ ] Local pack + install test confirms \`node_modules/mvlite/\` and the matching \`node_modules/mvlite-<platform>/\` appear after \`npm install movehat-*.tgz\` (run from a tmp dir)
- [ ] \`pnpm test:example\` shows "Starting mvlite..." instead of "Starting local Movement node..." (proves auto-detection finds the new dep)
- [ ] \`pnpm test:example\` still passes when \`node_modules/mvlite*\` is manually deleted (fallback regression check)

## Blocker

mvlite 0.1.0 must be published to npm before this PR can merge. Tracked in https://github.com/gilbertsahumada/mvlite/pull/3.

gilbertsahumada:
> **Still blocked on movelite 0.1.0 publication.** Do not merge until `npm view movelite version` returns `0.1.0`. (Package renamed from `mvlite` — npm rejected that name as too similar to an unrelated `mv-lite`.)

## Summary

Aligns movehat's auto-spawn integration (shipped in #298/#299) with the upstream rename `mvlite` -> `movelite`, declares the optionalDependency, and fixes an auth-default mismatch.

### Rename (mvlite -> movelite)
- `MvliteManager` -> `MoveliteManager` (file + class)
- `findMvliteBinary` -> `findMoveliteBinary`; platform packages `mvlite-*` -> `movelite-*`; resolved binary `bin/mvlite` -> `bin/movelite`; `MVLITE_PATH` -> `MOVELITE_PATH`; `which mvlite` -> `which movelite`
- `LocalTestOptions.useMvlite` -> `useMovelite`
- `optionalDependencies`: `movelite: ^0.1.0`
- CHANGELOG `[Unreleased]` entries

### Auth fix (--no-auth)
movelite 0.1.0 generates a random auth token by default and gates `/mint` behind `x-movelite-token`. `MoveliteManager` now spawns with `--no-auth`, otherwise `fundAccounts()` would 401. movehat owns the process and binds to 127.0.0.1, so disabling local auth for the embedded spawn is correct.

## Verification (this branch)
- Tier 1: `pnpm check:example` passes.
- Unit: 560/560 pass.
- §6.2 Tier 2 (`pnpm test:example` against published movelite): deferred until movelite 0.1.0 is on npm. Expected: boot switches to "Starting movelite..." <1s.

## Tier 2 / Tier 3 status
- `pnpm test:smoke`: N/A — no template/bin changes; metadata + node-manager rename only.
- `pnpm test:example` against published movelite: pending pre-merge.
- `pnpm test:e2e`: runs in the develop -> main batch.

## Self-review (per §8)
Pending pre-merge after the blocker clears.

Closes #300.
Tracks gilbertsahumada/movelite (rename + Apache-2.0 refactor merged there).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,6 +16,14 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   to the full Movement node if mvlite is not installed. Opt out with
   `useMvlite: false` in `LocalTestOptions`. Closes #192.
 
+- `mvlite` declared as an `optionalDependency`. `npm install movehat`
+  now also downloads the matching `mvlite-<platform>` binary
+  (via mvlite's own optional dependencies on its platform packages),
+  so the auto-spawn path above activates without a separate manual
+  install step. Supported platforms: macOS arm64/x64, Linux x64/arm64.
+  Unsupported platforms (e.g. Windows) install movehat cleanly and
+  fall back to the Movement node. Closes #300.
+
 ### Security
 
 - Runtime validation at every fork API and storage boundary. All
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,16 +9,16 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Added
 
-- mvlite auto-spawn integration. `Harness.createLocal()` and
+- movelite auto-spawn integration. `Harness.createLocal()` and
   `setupTestFixture()` automatically detect and use
-  [mvlite](https://github.com/gilbertsahumada/mvlite) if the binary
+  [movelite](https://github.com/gilbertsahumada/movelite) if the binary
   is available, reducing test boot time from ~15s to <1s. Falls back
-  to the full Movement node if mvlite is not installed. Opt out with
-  `useMvlite: false` in `LocalTestOptions`. Closes #192.
+  to the full Movement node if movelite is not installed. Opt out with
+  `useMovelite: false` in `LocalTestOptions`. Closes #192.
 
-- `mvlite` declared as an `optionalDependency`. `npm install movehat`
-  now also downloads the matching `mvlite-<platform>` binary
-  (via mvlite's own optional dependencies on its platform packages),
+- `movelite` declared as an `optionalDependency`. `npm install movehat`
+  now also downloads the matching `movelite-<platform>` binary
+  (via movelite's own optional dependencies on its platform packages),
   so the auto-spawn path above activates without a separate manual
   install step. Supported platforms: macOS arm64/x64, Linux x64/arm64.
   Unsupported platforms (e.g. Windows) install movehat cleanly and
--- a/packages/movehat/package.json
+++ b/packages/movehat/package.json
@@ -70,6 +70,9 @@
     "prompts": "^2.4.2",
     "tsx": "^4.7.0"
   },
+  "optionalDependencies": {
+    "mvlite": "^0.1.0"
+  },
   "devDependencies": {
     "@types/js-yaml": "^4.0.9",
     "@types/node": "^24.10.1",
--- a/packages/movehat/package.json
+++ b/packages/movehat/package.json
@@ -71,7 +71,7 @@
     "tsx": "^4.7.0"
   },
   "optionalDependencies": {
-    "mvlite": "^0.1.0"
+    "movelite": "^0.1.0"
   },
   "devDependencies": {
     "@types/js-yaml": "^4.0.9",
--- a/packages/movehat/src/helpers/index.ts
+++ b/packages/movehat/src/helpers/index.ts
@@ -26,7 +26,7 @@ export { AccountManager } from "../core/AccountManager.js";
 export type { StoredAccount } from "../core/AccountManager.js";
 export { LocalNodeManager } from "../node/LocalNodeManager.js";
 export type { LocalNodeOptions, LocalNodeInfo } from "../node/LocalNodeManager.js";
-export { MvliteManager, findMvliteBinary } from "../node/MvliteManager.js";
+export { MoveliteManager, findMoveliteBinary } from "../node/MoveliteManager.js";
 export { setupLocalTesting } from "./setupLocalTesting.js";
 export type { LocalTestingContext } from "./setupLocalTesting.js";
 export {
--- a/packages/movehat/src/helpers/setupLocalTesting.ts
+++ b/packages/movehat/src/helpers/setupLocalTesting.ts
@@ -6,7 +6,7 @@ import { ForkManager } from "../fork/manager.js";
 import { ForkServer } from "../fork/server.js";
 import { LocalNodeManager } from "../node/LocalNodeManager.js";
 import type { LocalNodeInfo } from "../node/LocalNodeManager.js";
-import { MvliteManager, findMvliteBinary } from "../node/MvliteManager.js";
+import { MoveliteManager, findMoveliteBinary } from "../node/MoveliteManager.js";
 import type { NodeProvider } from "../node/NodeProvider.js";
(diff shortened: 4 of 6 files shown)


gilbertsahumada (commented):
## Self-review (per §8)

11-line metadata change. Did the structured re-read anyway. Findings:

### 🔴 Blockers

None.

1. **Verify pnpm-workspace install actually honors `optionalDependencies` declared inside a workspace package.** The Tier 2 plan ("after publish, `pnpm test:example` should switch to mvlite") rests on the assumption that `pnpm install` at the repo root will install `mvlite` into the workspace's hoisted `node_modules/` because it appears under `packages/movehat/package.json`'s `optionalDependencies`. pnpm does honor this in practice, but worth confirming with `pnpm install --filter movehat` + `ls node_modules/.pnpm/ | grep mvlite` once mvlite@0.1.0 is on npm. If it doesn't hoist, `findMvliteBinary()`'s `require.resolve` won't see it and the test won't actually flip to the mvlite path — making the post-blocker verification silently meaningless.

2. **No automated regression gate for "mvlite-was-actually-found".** Tier 1 typechecks the surface, Tier 2 runs the example, but neither asserts which node backend was used. After unblock, the example test passes whether mvlite spawned (boot <1s, "Starting mvlite…") or the Movement node spawned (boot 51s, "Starting local Movement node…"). A future mvlite version bump or rename could silently regress us back to the slow path with all gates green. Worth filing a follow-up issue: "Add assertion to `pnpm test:example` that fails if the boot path doesn't match the expected backend" — could grep the spawned process command or check wall-clock against a threshold.

### 🟢 Nits / follow-ups

- **`[Unreleased]` has two `### Added` sections** (lines 10 and 55). Pre-existing — not introduced by this PR. Mentioning so the release PR consolidates them when cutting 0.2.8.
- **README doesn't mention mvlite anywhere.** Once unblocked + published, the install section should call out the new sub-second boot. Not blocking — fits a small docs-sync follow-up sub-PR.
- The `^0.1.0` caret on a pre-1.0 dep means npm treats it as `>=0.1.0 <0.2.0`. When mvlite cuts 0.2.0 (likely a breaking SDK-contract change), movehat needs a deliberate bump. Already captured in the commit message; no action needed here.

### Pre-merge gate (when blocker clears)

1. `npm view mvlite version` returns `0.1.0`.
2. `pnpm install` at the repo root.
3. `ls node_modules/.pnpm/ | grep mvlite` shows the platform package.
4. `pnpm test:example` output starts with "Starting mvlite…", finishes 9/9 in <30s.
5. Post resolution comment listing the 🟡 outcomes (verified or deferred to follow-up issue).

gilbertsahumada:
## Resolution of self-review

### 🟡 #1 — pnpm workspace install honors workspace-package optionalDependencies → **resolved (empirically verified)**

Test method: temporarily added `fsevents@^2.3.3` (known-published, tiny, harmless) alongside `mvlite` in `packages/movehat/package.json` `optionalDependencies`, ran `pnpm install --filter movehat`, inspected:

- `node_modules/.pnpm/` → contains `fsevents@2.3.3` ✓
- `packages/movehat/node_modules/fsevents` → present as symlink ✓
- `pnpm-lock.yaml` → `optionalDependencies: fsevents: ^2.3.3 → 2.3.3` block under the workspace package entry ✓
- mvlite (unpublished) → silently absent from lockfile, no error or warning ✓

Conclusion: pnpm honors workspace-package `optionalDependencies`. Once mvlite@0.1.0 is on npm, the same `pnpm install` from develop will hoist `mvlite` + the matching `mvlite-<platform>` exactly the way `fsevents` got hoisted just now. Reverted the temporary `fsevents` entry; final diff is unchanged from the original PR.

### 🟡 #2 — Backend assertion gate → **deferred to follow-up issue #302**

Filed https://github.com/gilbertsahumada/movehat/issues/302 with a DoD: assertion that fails `pnpm test:example` if the expected backend (mvlite vs Movement node) doesn't actually spawn, and if boot time exceeds the threshold for that backend. Not blocking this PR (which is only metadata) but tracking explicitly so a future mvlite version bump or rename can't silently regress us back to the slow path with all gates green.

### 🟢 Nits

- **Duplicate `### Added` in `[Unreleased]`** — pre-existing CHANGELOG hygiene issue, not introduced here. Deliberately left for the release PR that cuts 0.2.8 (per §3 surgical-changes: don't fix unrelated state in this PR).
- **README mvlite mention** — deferred to a docs-sync sub-PR after `pnpm test:example` confirms the fast-boot path is live. Adding it now would document a behavior that doesn't activate until publish.
- **Caret `^0.1.0` semantics on pre-1.0 dep** — non-action: already captured in the commit body. When mvlite cuts 0.2.0, movehat needs a deliberate bump.

### Pre-merge gate (unchanged from review)

1. `npm view mvlite version` returns `0.1.0`
2. `pnpm install` at repo root
3. `ls node_modules/.pnpm/ | grep mvlite` shows platform package
4. `pnpm test:example` starts with `Starting mvlite...`, finishes 9/9 in <30s

gilbertsahumada (commented):
## Self-review (per §8)

### 🔴 Blockers
None.

### 🟡 Worth noting
1. **`pnpm test:example` is NOT a movelite gate.** The example uses the shared-node pattern (`tests/setup.ts` starts one `LocalNodeManager` and each spec passes `{ localNode: getSharedNode() }`), which takes the `if (options.localNode)` branch and bypasses movelite entirely. So this PR's movelite path is not exercised by the example. Verified movelite directly instead (see below). Tracked: #302 (add a real movelite-backed assertion).
2. **Example run was flaky on the Movement-node path** (4 failing: `before all` 60s timeouts + undefined-context teardown). This path is untouched by this PR — the `localNode` branch is byte-identical. One spec (`message`) passed against the same shared node, confirming dist + that path work. Investigating the flakiness as a separate follow-up; not a regression from this change.

### 🟢
- pnpm two-hop resolution fix is strictly more correct (works under npm AND pnpm); without it auto-spawn silently never fired under pnpm/yarn.
- `--no-auth` spawn is correct for the embedded case (movehat owns the 127.0.0.1 process).

## Verification
- Tier 1 (`pnpm check:example`): PASS.
- Unit (`pnpm test`): 560/560 PASS.
- movelite integration (end-to-end, against published movelite@0.1.0):
  - `findMoveliteBinary()` resolves the installed platform binary.
  - `MoveliteManager` spawn → ready (<1s) → `fundAccount` (no 401, --no-auth works) → clean stop.
  - movelite REST/SDK surface: 11/11 in the movelite repo.
- §6.2 `pnpm test:example`: Movement-node path flaky this run (unrelated to this PR; investigation to follow). movelite path verified directly above.
- `pnpm test:smoke` / `test:e2e`: N/A here — runs in the develop → main batch.

Merging to develop (CR-skip per develop-targeted workflow). Flakiness investigation + #302 follow up.

Closes #300.

PR #306, opened 2026-05-31
fix: deploy via SDK for movelite + backend-assertion gate

### LINKED ISSUE #302: test:example should fail when expected backend doesn't spawn (mvlite vs Movement node)
## Problem

After PR #298 (auto-spawn) + #301 (optionalDependency), \`pnpm test:example\` succeeds in two very different configurations:

- **Fast path**: mvlite spawned, boot <1s, log starts with \`Starting mvlite...\`
- **Slow path**: Movement node spawned, boot ~15s, log starts with \`Starting local Movement node...\`

Both run identical fixtures, both reach \`9 passing\`. The 51× boot-time difference is invisible to any green gate. That means a future mvlite version bump, an internal API change in mvlite, or a rename of \`findMvliteBinary\`'s lookup path could silently regress us back to the slow path with **all Tier 1 / Tier 2 / Tier 3 gates green**.

The fast-boot promise made to users in the CHANGELOG entry for #192 / #300 only holds if it's automatically enforced.

## Definition of Done

- [ ] An assertion exists in \`scripts/e2e-local.sh\` or in a new \`scripts/test-example-strict.sh\` (whichever fits the existing harness shape) that runs \`pnpm test:example\` and fails if either:
  - (a) The first \`Starting...\` log line doesn't match \`MOVEHAT_EXPECT_BACKEND\` (default: \`mvlite\` when the binary is detected, \`movement-node\` otherwise), OR
  - (b) Total wall-clock exceeds a threshold (\`MOVEHAT_EXPECT_BOOT_MAX_MS\` env var, default 5000 when mvlite is expected, 90000 when Movement node is expected).
- [ ] CI workflow runs this in the mvlite-expected mode (after \`npm install mvlite\` is wired up).
- [ ] CI workflow runs this in the movement-node-expected mode (with \`useMvlite: false\` forced, or with mvlite-* packages stripped from node_modules).
- [ ] Failure output is actionable: prints which backend actually ran, the boot time, and which env var was violated.

## Why now

Filed as follow-up to the §8 self-review on PR #301 (https://github.com/gilbertsahumada/movehat/pull/301). Defers cleanly behind both #298 (already shipped) and #301 (currently blocked on mvlite 0.1.0 publish). Should be picked up alongside the post-publish \`develop -> main\` batch closeout, or earlier if any mvlite-related test failure is observed.

## Out of scope

- Per-test assertions (e.g. \`expect(boot).lessThan(X)\` inside the mocha specs). The harness-level check is sufficient and avoids polluting the test files.
- Cross-platform CI matrix expansion. Use whichever single platform CI currently exercises.

### LINKED ISSUE #305: movelite autoDeploy fails: Movement CLI 'move publish' can't consume movelite REST responses
## Problem

`setupTestFixture` / `setupLocalTesting` auto-spawn movelite (0.2.8) for fast local testing, then `autoDeploy` publishes the test modules. Publishing goes through `Publisher.deploy` → `runCli("movement", ["move","publish", ...])`. The Movement CLI's REST client cannot consume movelite's responses, so every autoDeploy against movelite fails:

```
move publish --url http://127.0.0.1:8090/v1 ...
"Error": "API error: Unknown error Failed to build State from headers due to
 missing values in response. Chain ID: None, Version: None, Timestamp: None..."
```

This means the 0.2.8 fast-boot path is broken end-to-end: movelite spawns, but no module can be deployed onto it via the CLI, so `setupTestFixture` throws `Module "<x>" was not deployed`.

## Root cause

The Movement CLI is the only strict consumer. Verified empirically against the **published** `movelite@0.1.0` binary:

| Consumer | Result |
|---|---|
| movelite VM (publish_package_txn) | works |
| `@aptos-labs/ts-sdk` publishPackageTransaction + waitForTransaction | works |
| Movement CLI `move publish` | fails ("Failed to build State from headers") |

movehat already depends on `@aptos-labs/ts-sdk`. The bounded fix is to publish via the SDK when the backend is movelite, keeping the CLI path for real Movement nodes / forks / testnet.

## Definition of Done

- [ ] `Publisher.deploy` publishes via `aptos.publishPackageTransaction` (SDK) when an internal `sdkPublish` flag is set; CLI path unchanged otherwise.
- [ ] `setupLocalTesting` sets `sdkPublish` when `localNode instanceof MoveliteManager`.
- [ ] The CLI build step emits `package-metadata.bcs` (`--save-metadata`) on the SDK path so the artifacts can be read.
- [ ] Proof: `MOVEHAT_USE_MOVELITE=1 movehat run scripts/backend-probe.ts` deploys `counter` and reaches `BACKEND-PROBE: ready`.
- [ ] Proof: full `setupTestFixture(["counter"],["deployer"])` + `counter.call(deployer,"increment",[])` + `counter.view("get",[addr])` returns `1` through movelite.
- [ ] No regression: `pnpm --filter movehat test` (560) green; `pnpm test:example` (9, real Movement node, CLI path) green; `pnpm typecheck:example` clean.

gilbertsahumada:
## Summary

The 0.2.8 fast-boot path spawned movelite but **could not deploy onto it**. `autoDeploy` publishes through the Movement CLI (`move publish`), whose REST client cannot consume movelite's responses:

```
"Error": "API error: Unknown error Failed to build State from headers due to
 missing values in response. Chain ID: None, Version: None, Timestamp: None..."
```

So every `setupTestFixture` against movelite threw `Module "<x>" was not deployed`. The fast-boot feature was broken end-to-end as shipped.

### Root cause (verified against published `movelite@0.1.0`)

| Consumer of movelite | Result |
|---|---|
| movelite VM (`publish_package_txn`) | works |
| `@aptos-labs/ts-sdk` publish + waitForTransaction | works |
| Movement CLI `move publish` | fails ("Failed to build State from headers") |

The Movement CLI is the only strict consumer. movehat already depends on the SDK.

### Fix

- `Publisher` publishes via `aptos.publishPackageTransaction` when the backend is movelite; the Movement CLI path is unchanged for real nodes, forks, and testnet (extracted verbatim into `publishViaCli`).
- The CLI build step gains `--save-metadata` on the SDK path so `package-metadata.bcs` exists to read.
- `setupLocalTesting` sets the internal `sdkPublish` flag when `localNode instanceof MoveliteManager`.

### Backend-assertion gate (Closes #302)

`scripts/assert-backend.sh` (`pnpm assert-backend`) drives the auto-spawn path via a dedicated probe and fails if the wrong local-test backend started — guarding the fast-boot promise against a silent regression to the slow Movement node. Identity (which start line appeared) is the primary guard; a fixture-ready time ceiling corroborates it.

## Verification

| Check | Result |
|---|---|
| `pnpm build:movehat` | clean |
| `pnpm --filter movehat test` | **560 passing** |
| `pnpm typecheck:example` | clean |
| `pnpm test:example` (real Movement node, **CLI path** regression) | **9 passing** (46s) |
| `MOVEHAT_EXPECT_BACKEND=movelite pnpm assert-backend` | **pass** (identity=movelite, 6.3s) |
| Full loop through movelite: deploy + `increment` + `view("get")` | returns **1** |

Tier-2: `test:smoke` / `test:e2e` not run (no change to packaging or templates; `bin/**` and `src/templates/**` untouched). `movement-node` gate mode not run here (~90s, needs local node) — covered transitively by `test:example` exercising the same CLI path.

## Notes

- No movelite republish needed: the SDK path works against the **published** `movelite@0.1.0` binary that users already install.
- `movehat run` (standalone) does not auto-spawn movelite; the only affected path is the test-fixture `autoDeploy`, which this PR fixes.

## Docs (§6.4)

- `guides/testing.mdx`: "Fast local boot with movelite" subsection (auto-pull, fallback, `useMovelite: false`).
- `getting-started/installation.mdx`: optional-dependency note.

CHANGELOG `[Unreleased]` updated (§11).

Closes #305
Closes #302


gilbertsahumada (commented):
## Self-review (§8)

Re-read the full diff. Findings:

### 🔴 Blockers
None.

### 🟡 Worth noting (deferred with rationale)

- **Module publish order** — `publishViaSdk` reads `bytecode_modules/*.mv` with a plain filename `.sort()`. This is the canonical aptos-ts-sdk pattern and the AptosVM bundle-verifies the package as a unit (order-tolerant), proven end-to-end for the 4-module counter package. A package whose modules require a strict non-alphabetical publish order is theoretically possible but not how the Move bundle publish works. Deferring; would surface immediately as a publish failure, not silent corruption.
- **Single-package assumption** — `pkgDirs.length !== 1` throws if `build/` holds more than one top-level package-metadata dir. movehat's model is one package per `moveDir`, so this matches reality and fails loudly (not silently) if violated. Deferring.
- **`movement-node` gate mode not run in this PR** (~90s, needs a local node). The same CLI publish path it would exercise is covered by `test:example` (9 passing against a real Movement node). Deferring the explicit gate run.
- **SDK-path error redaction** — the CLI path runs through `runCli`'s key redaction; the SDK path signs in-process and its errors carry tx data, not the private key, so there is no key-leak surface to redact. No action.

### 🟢 Nits
- `aptos` is passed to `Publisher.deploy` unconditionally and only read on the SDK path. Harmless; keeps the call site uniform.

### Pre-merge checklist
- [x] Self-review posted
- [x] 🟡 items resolved-or-deferred with rationale (all deferred, reasons above)
- [x] Tier-2 results in PR body (`test:example` pass; smoke/e2e N/A — no packaging/template/bin change)
- [ ] `mergeable` CLEAN — to confirm before merge

Note: targets `develop`; merge does not publish (only the develop→main batch does). Holding merge for your review.

gilbertsahumada:
## Code review (`/review`)

### Overview
Fixes the 0.2.8 movelite deploy break: `autoDeploy` published through the Movement CLI, which can't consume movelite's REST responses. `Publisher.deploy` now branches after the (shared) build step — `publishViaCli` for real nodes/forks/testnet, `publishViaSdk` (`@aptos-labs/ts-sdk publishPackageTransaction`) for movelite. `setupLocalTesting` sets the internal `sdkPublish` flag when `localNode instanceof MoveliteManager`. Adds the `assert-backend` gate (#302) + docs.

### Code quality
- The CLI publish logic is extracted **verbatim** into `publishViaCli` — confirmed behavior-preserving by `test:example` (9 passing against a real node) and `runtime.ts` staying at 100% coverage. Good refactor: the diff reads as a clean move plus one new sibling method.
- The strategy branch in `deploy()` is small and readable; the shared build + post-publish deployment-record/`PostPublishError` tail correctly wraps both paths, so redeploy-guard and bookkeeping semantics are identical regardless of backend.
- Internal `sdkPublish`/`aptos` plumbing is minimal and matches the existing `adapter` "test-only internal" precedent on the same options object.
- **Security note (positive):** the SDK path signs in-process and never writes a temp private-key file, so it sidesteps the whole `keyFilePath` / SIGINT-cleanup leak surface the CLI path has to manage.

### Findings

**🟡 The new `publishViaSdk` path has no unit-test coverage.** `Publisher.ts` dropped to 62.76% line coverage; the SDK branch (`publishPackageTransaction` → sign → submit → `waitForTransaction`, plus the build-artifact reads) is exercised only by the Tier-2 `assert-backend` probe, which is local-only and deferred from CI. Global coverage is 76.27% (> 70% floor) so **CI stays green** and this is not blocking — but a unit test that mocks `aptos.publishPackageTransaction`/`submit`/`waitForTransaction` and the `fs` reads (happy path + the two throw paths: missing `aptos`, `pkgDirs.length !== 1`) would protect a now-critical path that currently has zero automated coverage. Recommend as a fast follow.

**🟢 `pkgDirs.length !== 1` assumes a single compiled package** under `build/`. Matches movehat's one-package-per-`moveDir` model and fails loudly (not silently) if violated. Fine as-is; worth revisiting if multi-package workspaces are ever supported.

**🟢 Module order** is a filename `.sort()` of `bytecode_modules/*.mv`. Correct because the AptosVM publishes the package as a bundle and verifies it as a unit (order-independent) — proven end-to-end for the 4-module counter package. A one-line comment stating that invariant would help future readers.

**🟢 `readdirSync(modulesDir)`** throws a raw ENOENT if `bytecode_modules` is somehow absent after a successful build; caught by `deploy()`'s outer handler and surfaced as `Failed to publish module: ...`. Acceptable.

### Risk / verification
- No movelite republish required — verified against the **published** `movelite@0.1.0` binary (not just a local build).
- `movehat run` (standalone) does not auto-spawn movelite, so the only affected path is the fixture `autoDeploy` this PR fixes — blast radius is contained.
- Full loop confirmed through movelite: deploy + `increment` + `view("get")` → `1`.

**Verdict:** solid, well-scoped, thoroughly verified. One non-blocking coverage gap (SDK path unit test) worth a follow-up. No correctness or security concerns.

gilbertsahumada:
## Resolution — coverage gap closed

The single 🟡 from the review (no unit coverage for `publishViaSdk`) is now addressed in `dc1e88a`.

Added 3 unit tests in `deployContract.test.ts` (`Publisher — SDK publish path`):
- **Happy path** — mocks the SDK client and pre-stages the compiled artifacts; asserts only `move build` (with `--save-metadata`) reaches the CLI, publish goes through the SDK, metadata + module bytecode are passed in sorted order, and `sign`/`submit`/`waitForTransaction` run with the hash recorded + persisted.
- **Missing `aptos`** — throws `sdkPublish requires an Aptos client`.
- **No compiled package** — throws `Expected exactly one compiled package`.

Result:
- `pnpm --filter movehat test`: **563 passing** (was 560)
- `Publisher.ts` line coverage: **62.76% → 89.36%**
- Global coverage: 77.15% (> 70% floor)

The SDK path now has automated CI coverage, not just the Tier-2 probe. No other findings outstanding.

PR #310, opened 2026-06-01
ci: movelite backend gate on Linux + Node 24 opt-in

gilbertsahumada:
## What

- **#3 — movelite gate in CI:** a step in the `e2e-tests` job runs `MOVEHAT_EXPECT_BACKEND=movelite pnpm assert-backend` on ubuntu. First time the example's deploy loop runs on Linux CI — the full Movement node can't start there (MintFunder bug, hence `MOVEHAT_SKIP_LOCAL_NODE`), but movelite can. Reuses the job's Movement CLI (for the `movement move build` compile step) instead of duplicating ~80 lines.
- **#4 — Node 24 opt-in:** `FORCE_JAVASCRIPT_ACTIONS_TO_NODE24=true` on `ci.yml` / `publish.yml` / `docs-deploy.yml`. GitHub's recommended opt-in ahead of the 2026-06-16 forced switch — no action-version change, so no breaking-change risk to the publish pipeline.
- **Cross-platform lockfile:** `pnpm.supportedArchitectures` (darwin + linux, x64 + arm64) + `pnpm update movelite` so the lockfile carries `movelite-linux-x64`. Without this the gate fell back to the Movement node and timed out — the lockfile's `movelite@0.1.0` snapshot predated the Linux packages' publication.

## Verification

Dispatched `ci.yml` on this branch (run 26766277272) — **all jobs green**, including:
```
movelite backend gate (#302): success
  ✓ Backend matched: 'movelite'
  ✓ Boot within budget: 6355ms <= 20000ms
```
Every job ran with `FORCE_JAVASCRIPT_ACTIONS_TO_NODE24` — no Node 20 deprecation warnings, nothing broke.

Tracks #302.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -12,6 +12,13 @@ on:
     branches: [ main ]
   workflow_dispatch:
 
+# Opt the pinned JS actions into the Node 24 runtime ahead of GitHub's
+# 2026-06-16 forced switch (the actions still run on Node 20 otherwise and
+# emit deprecation warnings). This is GitHub's recommended opt-in; it does not
+# change the action versions, so there is no breaking-change risk.
+env:
+  FORCE_JAVASCRIPT_ACTIONS_TO_NODE24: true
+
 jobs:
   # Build and type check
   build:
@@ -151,7 +158,7 @@ jobs:
     name: E2E Tests
     runs-on: ubuntu-latest
     needs: [build, unit-tests]
-    timeout-minutes: 5
+    timeout-minutes: 12
 
     steps:
       - uses: actions/checkout@34e114876b0b11c390a56381ad16ebd13914f8d5  # v4.3.1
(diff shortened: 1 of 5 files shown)


gilbertsahumada:
## Code review (`/review`)

### Overview
Wires the movelite backend gate into Linux CI (#302), opts the pinned JS actions into Node 24, and makes the lockfile cross-platform so `movelite-linux-x64` actually installs in CI. Verified green on dispatch run 26766277272 (`movelite backend gate: success`, Backend matched: movelite, boot 6.3s).

### Findings

**🟡 Lockfile carries unrelated transitive patch bumps.** Beyond the intended movelite additions, the `pnpm update` re-resolution also bumped `@types/node` 24.10.1→24.12.4, `@types/http-cache-semantics` 4.0.4→4.2.0, and `pump` 3.0.3→3.0.4 (transitive dev/type deps under `got`). These are patch-level, dev-only/transitive, and CI is green with them — harmless, but they're scope creep for a PR whose intent is the movelite platform entries. Acceptable to keep for lockfile consistency; flagging for transparency. If you'd rather keep the diff surgical, the lockfile can be regenerated isolating just the movelite block — but it's not worth a re-run for patch bumps.

**🟢 `MOVEMENT_RPC_URL` on the movelite gate step is likely unnecessary.** The gate runs movelite locally; the `movement move build` compile step doesn't hit an RPC, and the deploy targets movelite, not testnet. Harmless (mirrors the sibling e2e steps), just not load-bearing here.

**🟢 The gate adds a network dependency** — the example compile downloads the aptos-framework git deps. Same flakiness surface as the existing `test:example` compile; not new risk, worth keeping in mind if CI gets flaky.

### Correctness / verification
- `supportedArchitectures` (darwin+linux × x64+arm64) is the correct pnpm mechanism for a cross-platform lockfile; `movelite-linux-x64` now carries `os: [linux]` / `cpu: [x64]` + integrity, so a frozen install on Linux installs it and skips it elsewhere.
- The movelite snapshot correctly omits `movelite-darwin-x64` (still unpublished) — no dangling reference.
- `FORCE_JAVASCRIPT_ACTIONS_TO_NODE24` is GitHub's documented opt-in; no action-version change, so the publish pipeline is untouched. Every job ran with it, green, no Node 20 warnings.
- e2e timeout 5→12 min is appropriate given the added compile + deploy.

**Verdict:** solid and actually verified in CI (the one thing that couldn't be tested locally). One 🟡 lockfile-churn note, non-blocking.

PR #319, opened 2026-06-11
feat: counted -v verbosity levels + getVerbosityLevel (traces PR1a)

gilbertsahumada:
Tracks #315. Closes #316. First of three sub-PRs for the movelite transaction-trace renderer.

## Summary

Foundational verbosity infrastructure — makes a numeric verbosity level (0–4) available so later PRs can gate trace rendering on it. **No trace rendering in this PR.**

- `cli.ts`: `-v/--verbose` becomes a **counted** Commander option (`-v`..`-vvvv` → 1..4). preAction keeps setting `MOVEHAT_VERBOSE=1` for level ≥ 1 (back-compat with `isVerbose()` + its ~10 callers) and additionally sets `MOVEHAT_VERBOSITY=<level>`.
- `logger.ts`: new `getVerbosityLevel(): number` — reads `MOVEHAT_VERBOSITY` (clamped 0–4), falling back to the legacy `MOVEHAT_VERBOSE=1` (level 1). `isVerbose()` and its callers untouched.
- Docs: `cli/global-flags.mdx` (counted `-v`, L0–L4 table, `MOVEHAT_VERBOSITY`) + `guides/console-output.mdx` (verbosity-levels subsection).

## Why env-var propagation

`contract.call()` runs inside the spawned mocha/script subprocess, so the level must travel via an env var (`MOVEHAT_VERBOSITY`) — the same reason `isVerbose()` reads `MOVEHAT_VERBOSE`. In-process logger config would never reach `call()` under `movehat test`.

## How tested

- `pnpm --filter movehat test` — 569/569 (6 new `getVerbosityLevel` tests: env→level, clamp, non-numeric, back-compat, precedence).
- Verified Commander counts occurrences: `-vvvv`→4, `-v -v -v`→3.
- `pnpm check:example` (Tier-1) green. `pnpm build:docs` clean.
- Tier 2 (§6.2): `cli.ts` is `src/**` but the change is the flag-parsing surface only (no runtime/packaging behavior touched); `test:example`/`test:smoke` N/A. The counted-flag behavior is covered by the unit tests + the Commander check above.

## Note on forward doc links

The L2–L4 / `/docs/guides/traces` references resolve once PR1c lands. All three sub-PRs ship together in one `develop→main` batch (docs deploy only from main), so the live docs site is never inconsistent. `build:docs` passes with the forward links.

## Checklist

- [x] `pnpm test` green
- [x] CHANGELOG `[Unreleased]` (`### Changed`)
- [x] §8 self-review (below)

Issue #315, opened 2026-06-11
[traces] movelite-only Foundry-style -vvvv execution-trace renderer (meta)


gilbertsahumada:
## Meta-issue

Render Foundry-style transaction execution traces in movehat at raised verbosity (`-vv`..`-vvvv`), sourced from movelite's `POST /v1/transactions/trace?commit=true` endpoint (shipped in movelite 0.2.0). When the backend is movelite and verbosity >= 2, `MoveContract.call(...)` renders an indented, colored call tree (per-frame `module::function`, decoded args, type-args, gas, return values, events, storage ops, abort stack) and still returns the existing `{hash, success, vm_status}`.

Traces are inherently **movelite-only**: the real Movement node's REST API does not expose internal Move call frames. Node-backend degradation (surfacing the events/changes `contract.ts` currently discards) is a separate later effort, out of scope here.

## Sub-issues (each its own PR off `develop`)

- [ ] PR1a — Verbosity infrastructure: counted `-v` (-v..-vvvv), `MOVEHAT_VERBOSITY` env propagation, `getVerbosityLevel()`. `isVerbose()` and its callers untouched.
- [ ] PR1b — Trace types + HTTP client + `MoveContract` wiring (trace path gated on movelite + level>=2; `commit=true` is the sole commit).
- [ ] PR1c — Tree renderer (L2 events / L3 user-tree with framework-event bubbling / L4 full + abort) + docs (`guides/traces.mdx`, cross-links).

## Closes when

All three sub-PRs merge and a real counter `increment(5)` at `-vvvv` on movelite renders the call tree and still commits (verified via a follow-up `view()`).

PR #320, opened 2026-06-11
feat: trace types + client + MoveContract wiring (traces PR1b)

### LINKED ISSUE #317: [traces] PR1b — trace types + HTTP client + MoveContract wiring
Tracks #315. Gated on PR1a (`getVerbosityLevel`).

## Definition of Done

- NEW `packages/movehat/src/core/trace/types.ts`: TS interfaces mirroring movelite's serialized structs (`TraceResponse`, `CallNode`, `AbortInfo`, `TracedArg`, `TracedEvent`, `StorageOp`). `return` keyed literally; `gas` commented as INTERNAL VM units (not octas).
- NEW `packages/movehat/src/core/trace/client.ts`: `traceTransaction({aptos, rpcUrl, transaction, senderAuthenticator}): Promise<{response, elapsedMs}>` — `generateSignedTransaction` -> POST BCS to `${rpcUrl}/transactions/trace?commit=true` (rpcUrl already ends in `/v1`; `Content-Type: application/x.aptos.signed_transaction+bcs`; no Accept; no auth). Non-2xx -> throw. Measure `performance.now()` around the POST.
- `core/contract.ts`: optional 4th ctor param `traceRpcUrl?`; `getContract(..., traceRpcUrl?)` forwards it. In `call()`, branch `if (this.traceRpcUrl && getVerbosityLevel()>=2)` -> trace path (commit=true is the SOLE commit; never also `submit.simple`), return `{hash:txn_hash, success, vm_status}`; else existing path UNCHANGED. `renderTrace` is a pass-through no-op stub in this PR.
- Thread `traceRpcUrl`: `InitRuntimeOptions` (runtime.ts) + `getContractHelper` forward; `setupLocalTesting.ts` sets it only when `localNode instanceof MoveliteManager` (`${nodeInfo.rpcUrl}/v1`). `createLive`/`createFork` never set it.
- Tests: parse `~/projects/mvlite/docs/trace-sample.json` into the types; a real movelite `increment` at level>=2 still commits (follow-up `view()` shows the increment) and returns the mapped result. Tier-2 (`test:example`, `test:smoke`) run.
- CHANGELOG `[Unreleased]` updated.

## Verify

`pnpm --filter movehat test`; counter `increment` at `-vv` on movelite returns `{success:true}` and the counter increments.

gilbertsahumada:
Tracks #315. Closes #317. Gated on PR1a (#319, `getVerbosityLevel`). Second of three sub-PRs.

## Summary

Plumbing for the execution-trace renderer — **no tree rendering yet** (that is PR1c).

- **NEW `core/trace/types.ts`** — TS mirror of movelite's `/v1/transactions/trace` JSON contract. Keys match the wire format (literal `return`; per-frame `gas` documented as internal VM units, distinct from the octa `gas_used`).
- **NEW `core/trace/client.ts`** — `traceTransaction(...)`: `generateSignedTransaction` → POST the BCS bytes to `${rpcUrl}/transactions/trace?commit=true`. `rpcUrl` already ends in `/v1`; BCS content type; no Accept; no auth (movelite runs `--no-auth`). `commit=true` is the sole commit. Non-2xx → throw. Times the POST for the renderer.
- **`contract.ts`** — optional 4th ctor param `traceRpcUrl`; `getContract` forwards it. `call()` branches: when `traceRpcUrl` is set AND `getVerbosityLevel() >= 2`, route through `traceTransaction` and map to `{hash, success, vm_status}`; otherwise the existing `submit.simple` + `waitForTransaction` path is unchanged. **Never also submits** (commit=true already committed).
- **`runtime.ts` / `setupLocalTesting.ts`** — thread `traceRpcUrl` down; set only on the movelite backend (`isMovelite` computed once, reused for `sdkPublish`). `createLive`/`createFork` never set it.

## How tested

- **Real movelite, end-to-end**: `increment` at `MOVEHAT_VERBOSITY=2` commits through the trace endpoint (`view` 0 → 1, proving `commit=true` mutated state); the default path (level 0) still commits unchanged on movelite. (Verified with a throwaway script, since `test:example` pins the shared Movement node and does not exercise movelite.)
- **Unit** (`pnpm --filter movehat test`, 574/574, 5 new): mock `fetch` + `generateSignedTransaction`, assert URL / headers (BCS content type, no Accept) / body / parse / non-2xx throw; fixture conformance (gas vs gas_used, u64-as-string args, storage `address` nullability). Fixture is a captured real `trace-sample.json`.
- **Tier 2 (§6.2)**: `pnpm test:smoke` pass; `pnpm test:example` 9 passing (shared-node path, confirms no regression to the normal submit flow). Tier-1 + this PR's `src` change covered.

## Notes

- A trace POST failure **throws** rather than falling back to `submit.simple` — with `commit=true` a fallback submit would risk a double-commit. A `success:false` (HTTP 200) is an abort, rendered/returned as `{success:false}`, not a fallback.
- `view()` is untouched (read-only, never traces).

## Checklist

- [x] `pnpm test` 574/574
- [x] Tier 2: `test:smoke` pass, `test:example` pass
- [x] CHANGELOG `[Unreleased]` (`### Internal`)
- [x] §8 self-review (below)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -33,6 +33,15 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Internal
 
+- Trace transaction types + HTTP client + `MoveContract` wiring (no renderer
+  yet). `core/trace/{types,client}.ts` model movelite's `/v1/transactions/trace`
+  contract and POST the BCS-signed transaction with `commit=true`. On movelite
+  at verbosity level >= 2, `contract.call(...)` now routes through that endpoint
+  (a single instrumented execution that also commits) instead of the normal
+  submit, preserving the `{hash, success, vm_status}` return shape; on the
+  Movement node and below level 2 the existing submit path is unchanged. The
+  call tree is rendered in a following change. Closes #317.
+
 - Bump the bundled `movelite` optional dependency from `^0.1.0` to `^0.2.0`.
   The 0.2.0 binary adds the `POST /v1/transactions/trace` endpoint that the
   upcoming Foundry-style trace renderer will consume; bumping the dependency
--- a/packages/movehat/src/core/contract.ts
+++ b/packages/movehat/src/core/contract.ts
@@ -5,6 +5,7 @@ import {
   type MoveFunctionId,
 } from "@aptos-labs/ts-sdk";
 import { logger } from "../ui/index.js";
+import { traceTransaction } from "./trace/client.js";
 
 export interface TransactionResult {
   hash: string;
@@ -16,7 +17,11 @@ export class MoveContract {
   constructor(
     private aptos: Aptos,
     private moduleAddress: string,
-    private moduleName: string
+    private moduleName: string,
+    // movelite RPC base (ending in `/v1`). Presence enables Foundry-style
+    // execution traces at verbosity level >= 2. Undefined on the Movement node
+    // (no trace endpoint) — calls use the normal submit path.
+    private traceRpcUrl?: string
   ) {}
 
   async call(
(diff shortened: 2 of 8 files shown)


gilbertsahumada (commented):
## §8 Self-review — PR1b (trace types + client + wiring)

### 🔴 Blockers
None.

### 🟡 Noted
- **`new Uint8Array(bytes)` copy on every traced call.** Required because the SDK returns `Uint8Array<ArrayBufferLike>`, which the fetch `BodyInit` type rejects under this TS lib config (the ArrayBuffer/SharedArrayBuffer split). The copy is trivial (a signed-txn payload) and only happens on the trace path (level >= 2). Documented inline.
- **Trace branch duplicates the `logger.success(... committed ...)` + `newline()` lines** from the normal submit path. Accepted: the two paths derive the status from different sources (`response.vm_status` vs `committedTxn`/`waitForTransaction`), so a shared helper would obscure more than it saves at this size.
- **Fixture is a captured `trace-sample.json`, not a canonical artifact.** The live movelite sample had already drifted from earlier notes (2 storage ops, `gas_used: 4`); the test asserts on structural invariants (op kinds, address nullability, gas vs gas_used ordering) rather than exact values, so it is robust to further drift.

### 🟢 Notes
- `void elapsedMs` marks the single seam PR1c fills (the `renderTrace` call); the branch is otherwise complete.
- The movelite-only gate is `this.traceRpcUrl && level >= 2` — `traceRpcUrl` is set exclusively on the movelite backend, so the real Movement node, `createLive`, and `createFork` can never reach the trace path. Fork is additionally guarded by the existing call-throwing proxy.

### Verification recap
574/574 unit (5 new, mocked fetch + SDK); real-movelite increment commits at level 2 (view 0->1) and the default path still commits; Tier-2 `test:smoke` + `test:example` green; Tier-1 clean.

Ready to merge.

PR #325, opened 2026-06-12
feat: parse movelite JSON error body in trace client

### LINKED ISSUE #324: traces: parse movelite's JSON error body for cleaner trace-POST failures
movelite 0.2.1 returns structured JSON error bodies `{message, error_code, vm_error_code}` (previously plain text). The trace client (`core/trace/client.ts`) currently includes the raw body verbatim in the thrown error, so a failed trace POST now surfaces a JSON blob instead of a readable message.

## Definition of Done

- On a non-2xx trace response, attempt to parse the body as JSON and surface `message` (plus `error_code` / `vm_error_code` when present); fall back to the raw text when the body is not JSON. Backward compatible with movelite 0.2.0's plain-text errors.
- Unit test: JSON error body -> clean message extracted; plain-text body -> raw text preserved.
- `pnpm --filter movehat test` green.

Tracks #315.

gilbertsahumada:
Closes #324. Tracks #315.

## Summary

movelite 0.2.1 returns structured JSON error bodies (`{message, error_code, vm_error_code}`) instead of plain text. The trace client (`core/trace/client.ts`) previously embedded the raw body verbatim, so a failed trace POST would surface a JSON blob. It now extracts the `message` (plus `error_code` / `vm_error_code` when present) and falls back to the raw text when the body is not JSON — backward compatible with movelite 0.2.0's plain-text errors.

Confirmed compatible otherwise: the client already sends `Content-Type: application/x.aptos.signed_transaction+bcs` (passes 0.2.1's BCS validation) and no `Accept` header (no 406). The success endpoint and the trace JSON contract are unchanged, so the renderer is untouched.

## How tested

- `pnpm --filter movehat test` green (2 new client tests: JSON error → `message (codes)` extracted; plain text → raw preserved).
- Tier-1 `check:example` + Tier-2 `test:smoke` green. Happy path unchanged (error branch only).

## CHANGELOG

Folded into the existing unreleased traces `### Added` bullet (the feature has not shipped yet) with `Closes #324`.

## Checklist

- [x] `pnpm test` green
- [x] Tier-1 + smoke green
- [x] CHANGELOG updated
- [x] §8 self-review (below)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -20,7 +20,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   footer. Traces are movelite-only (the Movement node does not expose the trace
   endpoint) and opt-in by verbosity, so the default test loop is unaffected. A
   render failure degrades to a warning and never fails a committed transaction.
-  New guide at `guides/traces.mdx`. Closes #318.
+  A failed trace request surfaces movelite's structured JSON error `message`
+  (with `error_code` / `vm_error_code` when present), falling back to raw text
+  for older movelite builds. New guide at `guides/traces.mdx`. Closes #318,
+  Closes #324.
 
 ### Changed
 
--- a/packages/movehat/src/core/trace/__tests__/client.test.ts
+++ b/packages/movehat/src/core/trace/__tests__/client.test.ts
@@ -71,6 +71,41 @@ describe("traceTransaction", () => {
       })
     ).rejects.toThrow(/400.*bad transaction/s);
   });
+
+  it("surfaces the `message` from a movelite JSON error body", async () => {
+    fetchMock.mockResolvedValue(
+      new Response(
+        JSON.stringify({
+          message: "transaction aborted",
+          error_code: "vm_error",
+          vm_error_code: 4004,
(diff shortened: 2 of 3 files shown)


gilbertsahumada (commented):
## §8 Self-review — trace client JSON error parsing

### 🔴 Blockers
None.

### 🟡 Noted
- The JSON shape `{message, error_code, vm_error_code}` comes from the movelite 0.2.1 report, not a captured real error (0.2.1 isn't published yet, and a failed trace POST is hard to trigger on demand). Mitigated: parsing is defensive — any non-JSON or missing-`message` body falls back to the raw text, so a shape mismatch degrades gracefully rather than throwing. When 0.2.1 lands, the lockfile-refresh step can spot-check a real error if one surfaces.

### 🟢 Notes
- Backward compatible: movelite 0.2.0's plain-text errors hit the `catch` and return unchanged. Both branches are unit-tested.
- Scope is the error path only — the success path, the response contract, and the renderer are untouched, so Tier-2 `test:example` (which exercises the happy path) is not the relevant gate here; `test:smoke` covered packaging.

Ready to merge.

PR #340, opened 2026-07-04
fix: shared test node prefers movelite so traces work in movehat test

### LINKED ISSUE #339: Shared test-node pattern pins LocalNodeManager, silently disabling movelite and execution traces in movehat test
## Problem

The canonical example (`examples/counter-example/tests/setup.ts`) and the `movehat init` template (`packages/movehat/src/templates/tests/setup.ts`) boot a shared `LocalNodeManager` in mocha root hooks and inject it into every fixture via `{ localNode: getSharedNode() }`.

Passing `localNode` takes the first branch of backend selection (`setupLocalTesting.ts:163`), so the movelite auto-spawn branch (`:173`) is never reached and `traceRpcUrl` is never wired (`:232-237`). Consequences:

- `movehat test --ts -vv/-vvv/-vvvv` never renders the Foundry-style call tree — only the degraded node trace at best.
- The suite boots the slow Movement node (~15s+) even where movelite (<1s boot) is installed; on machines where the Movement node is unresponsive the whole example suite times out (observed: 0 passing / 5 failing at 3min).
- The scaffold contradicts the docs (`guides/testing.mdx` says movelite selection "is automatic and requires no configuration").

Additionally, `LocalTestOptions.localNode` is typed as `LocalNodeManager`, so a `MoveliteManager` cannot even be injected without widening the public type.

## Definition of Done

- [ ] `LocalTestOptions.localNode` widened to `NodeProvider` (`packages/movehat/src/types/config.ts`); `NodeProvider` exported from `movehat/helpers`.
- [ ] `examples/counter-example/tests/setup.ts` prefers movelite (`findMoveliteBinary()` → `MoveliteManager`) with `LocalNodeManager` fallback; `pnpm test:example` passes on the movelite backend.
- [ ] `packages/movehat/src/templates/tests/setup.ts` ships the same movelite-preferring pattern (`movehat init` output matches example + docs).
- [ ] `guides/testing.mdx` documents the shared-node pattern concisely, including the "injecting `localNode` opts out of movelite auto-selection" caveat.
- [ ] `guides/networks-and-modes.mdx` stale "#235 tracks this; until that ships" paragraph refreshed (the pattern shipped).
- [ ] `CHANGELOG.md [Unreleased]` entry.
- [ ] Verified manually: `MOVEHAT_VERBOSITY=4` on the example suite renders the full movelite call tree.

## Follow-up (separate issue, out of scope here)

Implicit movelite-only shared-node singleton inside `setupLocalTesting` so user test files contain zero node-lifecycle code (Hardhat/Foundry-parity DX). Requires `MoveliteManager` lifecycle hardening (stream teardown + unref + exit/signal kill hooks + liveness check in `waitForReady`) and closing the `deployments/local` + shared `build/` dir races. Will delete `tests/setup.ts` from example and template.

gilbertsahumada:
## Summary

The example suite and the `movehat init` scaffold pinned `LocalNodeManager` as the shared root-hook test node. An injected `localNode` takes precedence over movelite auto-selection (`setupLocalTesting.ts:163`), so `traceRpcUrl` was never wired: `movehat test --ts -vv..-vvvv` never rendered the Foundry-style call tree, and every run paid the slow Movement-node boot (on this machine the suite timed out entirely: 0 passing / 5 failing at 3min).

This PR makes the shared-node pattern movelite-preferring everywhere it exists today and syncs the docs. One commit per file:

- `fix: widen LocalTestOptions.localNode to NodeProvider` — the public option only accepted `LocalNodeManager`, so a `MoveliteManager` could not be injected at all.
- `feat: export NodeProvider type from movehat/helpers` — lets scaffold code annotate the shared node.
- `fix: example shared test node prefers movelite so traces work` — `findMoveliteBinary()` → `MoveliteManager`, `LocalNodeManager` fallback.
- `fix: init template shared test node prefers movelite` — `movehat init` output now matches the example and the docs.
- `docs: concise shared-node note in testing guide` — documents the scaffold pattern + the "injecting `localNode` opts out of movelite auto-selection" caveat.
- `docs: refresh stale shared-node paragraph in networks-and-modes` — the root-hooks pattern shipped (#235); the paragraph said it hadn't.
- `docs(changelog): movelite-preferring shared node + NodeProvider export`

Follow-up (separate issue, out of scope): implicit movelite-only shared-node singleton inside the framework so user test files carry zero node-lifecycle code (Hardhat/Foundry-parity DX). Requires `MoveliteManager` lifecycle hardening first.

## Verification (§6.2 Tier 2 / §6.3)

- `pnpm build:movehat` + `pnpm typecheck:example` (Tier 1): **pass** (also enforced by pre-push hook)
- `pnpm test:example`: **pass** — 9 passing in 24s on the movelite backend (baseline before this fix: 0 passing, Movement-node timeout). Note: one run executed concurrently with `pnpm build:docs` flaked on the fixture's 60s timeout from CPU contention; a solo re-run passed clean.
- `pnpm test:smoke`: **pass** — pack, global install, CLI commands, `movehat init` (covers the template change)
- `pnpm test:e2e`: **skip** — not a develop→main batch nor a publish; `templates/scripts/**` untouched (only `templates/tests/`), so the §6.2 e2e:quick rule does not apply
- `pnpm assert-backend`: **skip** — backend-selection logic (`setupLocalTesting.ts`, `MoveliteManager.ts`, `Publisher.ts`) untouched
- `pnpm build:docs`: **pass**
- Manual trace check: `MOVEHAT_VERBOSITY=2/3/4` on the example suite renders events / user-module tree / full call tree respectively via movelite's trace endpoint

Closes #339

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,26 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Fixed
+
+- The example suite and the `movehat init` scaffold silently disabled movelite
+  — and with it the Foundry-style execution traces — in `movehat test`. Their
+  shared root-hook node (`tests/setup.ts`) pinned `LocalNodeManager`, and an
+  injected `localNode` takes precedence over movelite auto-selection, so
+  `traceRpcUrl` was never wired and `-vv..-vvvv` rendered no call tree while
+  every run paid the slow Movement-node boot. The shared node now prefers
+  movelite when its binary resolves (Movement node remains the fallback), so
+  scaffolded projects get sub-second boots and full traces out of the box.
+  Closes #339.
+
+### Changed
+
+- `LocalTestOptions.localNode` now accepts any `NodeProvider` (previously
+  `LocalNodeManager` only), and the `NodeProvider` interface is exported from
+  `movehat/helpers` — injecting a pre-started `MoveliteManager` into
+  `setupTestFixture` / `Harness.createLocal` is now expressible in user code.
+  Closes #339.
+
 ## [0.4.1] - 2026-06-15
 
 ### Fixed
--- a/examples/counter-example/tests/setup.ts
+++ b/examples/counter-example/tests/setup.ts
@@ -1,13 +1,25 @@
-import { LocalNodeManager } from "movehat/helpers";
+import {
+  LocalNodeManager,
+  MoveliteManager,
+  findMoveliteBinary,
+  type NodeProvider,
+} from "movehat/helpers";
 
-let sharedNode: LocalNodeManager | undefined;
+let sharedNode: NodeProvider | undefined;
 
 /**
- * Returns the shared local Movement node started by root hooks.
+ * Returns the shared local node started by root hooks.
  * Pass the result as `{ localNode: getSharedNode() }` to
  * Harness.createLocal() or setupTestFixture().
+ *
+ * Prefers movelite when its binary is available: only the movelite backend
+ * exposes the /transactions/trace endpoint that powers the Foundry-style
+ * execution traces you see at raised verbosity (`movehat test --ts -vvvv`).
+ * When movelite is not installed (e.g. a platform without a published
+ * binary) it falls back to the full Movement node, which still runs the
+ * tests but renders only the degraded event/state trace.
  */
-export function getSharedNode(): LocalNodeManager {
+export function getSharedNode(): NodeProvider {
   if (!sharedNode || !sharedNode.isRunning()) {
     throw new Error(
       "Shared node not available. " +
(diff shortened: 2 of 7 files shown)


gilbertsahumada (commented):
## Self-review (§8)

Re-read the full diff after all gates passed.

### 🔴 Blockers

None found.

None found. Verified explicitly:

- **Type widening is safe**: `setupWithLocalNode` only touches `isRunning()` / `getNodeInfo()` on an injected node — both on the `NodeProvider` interface. `Harness` already typed its field as `NodeProvider` internally; the public option was the only narrow point.
- **Docs anchor is real**: `testing#sharing-one-node-across-specs-optional` matches the generated id in the built HTML (`out/docs/guides/testing.html`).
- **Template semantics preserved**: fallback branch keeps `forceRestart: true` and the 60s hook timeout; only the preferred backend changed. `@ts-nocheck` retained, no type annotations introduced into the JS-flavored template.
- **Strictest-consumer check**: example compiles under `noUncheckedIndexedAccess` + `exactOptionalPropertyTypes` + `verbatimModuleSyntax` (`type` modifier used on the `NodeProvider` import).

### 🟢 Nits

1. `testing.mdx` links the example setup file as `/tree/main/...` — canonical for a file is `/blob/main/...`. GitHub redirects, so the link works; fixing for correctness.
2. `networks-and-modes.mdx` symptom line reads slightly awkward after dropping the "2-5 minutes" figure; acceptable as-is, leaving it.

Applying nit 1 as a follow-up commit on this branch; merging after it lands.

PR #350, opened 2026-07-07
fix: pre-publish gate no longer aborts on the happy path

gilbertsahumada:
## Summary

`scripts/pre-publish.sh` killed itself on a clean tree: `check()` inspected `$?`, which at else-branch call sites is the exit status of the just-failed `if` condition — so a clean tree, an absent version tag, or a reasonable package size each printed a false ✗ and `set -e` aborted precisely when everything was fine (0.5.0's gate ran manually because of this). One commit:

- `check()` replaced with explicit `check_ok` / `check_fail` markers (no `$?` reliance anywhere; both return 0 so failures accumulate to the existing summary, which already exits 1 when any check failed — the original design intent, restored).
- New `--non-interactive` flag (also honors `CI=true`): the two "Continue anyway?" prompts become immediate failures for automation.
- Local npm-credentials check downgraded to informational: the publish path is OIDC via `publish.yml` on GitHub Release; local login only matters for a manual `npm publish` fallback. The closing hint now names the real path (`gh release create vX.Y.Z`).

Out of scope (pre-existing, warning-only): the package-size parser assumes MB units; a kB-sized tarball would mis-warn.

## Verification (§6.2)

- **Happy path (the bug)**: full run to completion through all 10 sections — typecheck, build, smoke, install-E2E — summary `Passed: 17 / Failed: 0`, exit 0. (Run from this branch with a temporary uncommitted version bump so the version-tag check has an untagged version to validate, answering both prompts; count reconciles exactly — the 18th check is the clean-tree ✓ that a dirty run correctly skips.)
- **Non-interactive, clean tree, feature branch**: exits 1 citing the branch; the formerly-fatal `No uncommitted changes` line now prints as ✓ first.
- **Non-interactive, dirty tree**: exits 1 immediately citing uncommitted changes; `CI=true` triggers the same guard.
- Tier 2 (`test:example` / `test:smoke`): N/A as a gate for this diff — maintainer script, not in `packages/movehat/src/**` nor the published tarball — but `test:smoke` and the install-E2E ran anyway inside the happy-path verification above, both green.
- shellcheck: not installed locally, skipped (best-effort item).

Closes #348

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,19 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Internal
+
+- The pre-publish gate (`scripts/pre-publish.sh`) no longer aborts on the
+  happy path. Its `check()` helper inspected `$?`, which at else-branch call
+  sites is the exit status of the just-failed `if` condition — so a clean
+  tree, an absent version tag, or a reasonable package size each printed a
+  false failure and killed the script under `set -e` precisely when
+  everything was fine (0.5.0 ran the gate's substance manually because of
+  this). Checks now use explicit `check_ok`/`check_fail` markers that
+  accumulate to the summary. Also adds `--non-interactive` (honors `CI=true`)
+  for automation, and treats missing local npm credentials as informational
+  since the publish path is OIDC via `publish.yml`. Closes #348.
+
 ## [0.5.0] - 2026-07-07
 
 ### Fixed
--- a/scripts/pre-publish.sh
+++ b/scripts/pre-publish.sh
@@ -3,10 +3,19 @@ set -e
 
 # Pre-Publish Checklist Script
 # We Run this before publishing to npm
+#
+# Usage: bash scripts/pre-publish.sh [--non-interactive]
+#   --non-interactive (or CI=true): never prompt; any condition that would
+#   have asked "Continue anyway?" fails immediately instead.
 
 # Detect project root
 SCRIPT_DIR="$(cd "$(dirname "${BASH_SOURCE[0]}")" && pwd)"
 PROJECT_ROOT="$(cd "$SCRIPT_DIR/.." && pwd)"
+
+NON_INTERACTIVE=0
+if [[ "${1:-}" == "--non-interactive" ]] || [[ "${CI:-}" == "true" ]]; then
+    NON_INTERACTIVE=1
+fi
 
 echo "=========================================="
 echo "MoveHat Pre-Publish Checklist"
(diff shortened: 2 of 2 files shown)


gilbertsahumada (commented):
## Self-review (§8)

Re-read the full script diff after verification.

### 🔴 / 🟡

None. Verified specifically:
- No remaining `$?`-dependent call sites: every `check_ok`/`check_fail` is unconditional within its branch, and the three inline `CHECKS_FAILED=$((...))` blocks were converted to `check_fail` so the summary counters stay consistent.
- Counter arithmetic keeps the `VAR=$((VAR + 1))` assignment form — `((VAR++))` would return nonzero on the first increment and re-trip `set -e`.
- The hard-exit sites that should stay hard exits (version tag already exists, line ~97) still `exit 1` directly — accumulate-and-continue is only for checks where later sections remain meaningful.
- `--non-interactive` parses `$1` and `CI=true`; both prompts covered; interactive behavior byte-identical otherwise.
- Empirical: full 10-section run passes (17/17 applicable checks), both non-interactive guards fire correctly, and the npm-login downgrade was exercised for real (this machine is not logged in — old script would have failed the gate).

### 🟢

- `printf 'y'`-style piping into the prompts requires no newlines (`read -n 1`); noted here for future automation, no code change needed.

Ready to merge.

gilbertsahumada (commented):
## Post-merge /review (adversarial pass; PR was merged before this ran)

**Core fix verified sound**: no remaining code path where a benign condition aborts under `set -e` (all `(( ))` / command-substitution / pipeline sites audited — conditions are exempt contexts; the helpers end in assignment-form arithmetic, avoiding the `((VAR++))`-returns-1 trap). Counter integrity holds: every remaining inline ✗ is immediately followed by `exit 1`, so no run that reaches the summary can desync ✓/✗ lines from the counters. README section checks, final `gh release create` hint rendering, and the sections-7-10 cwd expectations (they intentionally run from `packages/movehat/`) all check out.

**Findings → follow-up (none block; PR's own changes are correct):**

- 🟡 pre-existing — **the package-size check (section 6) is fully inert**: it greps `"Unpacked size:"` but npm emits lowercase `npm notice unpacked size: …`, so `PACKAGE_SIZE` is always empty, bc prints a parse error to stderr, and `check_ok "Package size is reasonable"` fires unconditionally. Behind it: `awk '{print $3}'` would extract the wrong field even with the case fixed, and the `> 10` compare ignores kB/MB units. The check has never worked.
- 🟡 pre-existing — **invoking from any directory other than repo root aborts with a raw node stack trace**: `PACKAGE_VERSION` is read relative to invocation cwd; `$PROJECT_ROOT` is computed but never used to normalize. One-line fix: `cd "$PROJECT_ROOT"` right after computing it.
- 🟢 new in this PR — **arg parsing accepts only an exact `$1 == --non-interactive`**: a typo'd or displaced flag silently runs interactive mode instead of erroring "unknown argument".
- Noted, mitigated by this PR: non-tty invocation without the flag (and without `CI=true`) still dies at `read` without a diagnostic — fail-closed, acceptable; the flag is the documented remedy.

Follow-up issue filed with the three items.

Issue #341, opened 2026-07-04
Implicit shared movelite node in setupLocalTesting — zero node-lifecycle code in user tests (Hardhat/Foundry-parity DX)


gilbertsahumada:
## Motivation

Neither Hardhat nor Foundry ever ask a dev to manage node lifecycle in test code: Hardhat Network boots in-process automatically ("an in-process Hardhat Network node is started automatically... no changes to your tests or scripts"), and `forge test` executes on an embedded revm — Anvil is only for external consumers (frontends, cast, serving forks). Movehat's equivalent optimization (one node per mocha run) currently lives in **user-visible scaffold code** (`tests/setup.ts` + `getSharedNode()` + `{ localNode }` injection), shipped by `movehat init` and the counter example since #235 and made movelite-preferring in #340.

The framework should own this: the first `setupTestFixture()` / `Harness.createLocal()` in a process boots one movelite node, later fixtures reuse it, shutdown is automatic. User test files then contain zero node code, and the scaffold's `tests/setup.ts` disappears.

## Design (validated by adversarial review, 2026-07-04)

**Scope guard — the singleton engages only when ALL hold:**
- no explicit `localNode` passed (stays the highest-precedence escape hatch),
- `resolveUseMovelite()` is true and `findMoveliteBinary()` resolves,
- no backend-affecting options passed (`useMovelite: false`, custom node ports, etc. → fall through to today's owned spawn, `ownsNode=true`). Anything else would silently serve the wrong backend and break the `MOVEHAT_EXPECT_BACKEND=movement-node pnpm assert-backend` gate.

**Movelite-only.** The `LocalNodeManager` fallback keeps per-fixture own+stop: a process-lifetime Movement-node singleton would hold port 8080 (which the fork server also defaults to), hit the template's 30s mocha timeout with a 60s boot, and its `forceRestart` `rmSync` is destructive across parallel workers.

**Mechanics:** module-level memoization of the in-flight `start()` promise (dedupes concurrent first callers); consumers get `ownsNode=false`, so `teardown()` / `Harness.cleanup()` semantics need zero changes (verified: fixture teardown already no-ops on non-owned nodes).

## Prerequisites (each its own sub-PR, §5)

[...]

PR #354, opened 2026-07-09
fix: harden MoveliteManager lifecycle for long-lived nodes

gilbertsahumada:
## Summary

Sub-PR 1 of 3 for the implicit shared movelite node (#341): makes `MoveliteManager` safe to become a process-lifetime singleton. Two commits:

1. **Enabler** — `SpawnedProcess.unref()` on the child-process adapter (unrefs the child AND each stdio pipe: they are separately ref'd handles, so unref'ing only the child leaves the sockets holding the event loop open); exports `CRITICAL_NODE_OUTPUT` so both node managers share one critical-signal filter.
2. **Hardening** — `MoveliteManager` migrates to the injectable `ChildProcessAdapter` (raw `child_process.spawn` gone; only `execSync` remains, in `findMoveliteBinary`) and closes the lifecycle gaps:
   - **Undrained pipes**: stdout/stderr now flow through the section-9 verbosity filter (critical signals always surface; chatter behind `-v`), so a chatty node can no longer block on the 64KB pipe buffer. The stderr handler also keeps a bounded tail (20 lines x 500 chars) for boot-failure messages.
   - **No unref**: the spawn handle is unref'd once ready — an unstopped node no longer keeps the mocha event loop alive.
   - **No exit hooks**: a sync SIGKILL hook (registered via `process.on("exit")` plus the existing `cleanupCallbacks`/`ensureSignalHandler` registry) reaps the child on parent exit and SIGINT/SIGTERM. SIGKILL because the exit handler gets one synchronous turn with no supervisor left to escalate, and movelite is in-memory with nothing to flush; an orphan would keep port 8090 bound and break the next run's 8090 -> 8091 -> throw probe. Hooks deregister on `stop()` (only after `exited` resolves, so they stay armed through the SIGTERM grace window) and when the child exits on its own.
   - **No liveness check in `waitForReady`**: the loop now checks child exit before each probe and before honoring a ready response (a foreign node on the same port cannot mask a dead child). A boot failure reports immediately with exit code / signal / spawn-error detail plus the stderr tail instead of burning the 15s timeout.
   - **`stop()` on the captured handle**: kills the pre-existing bug where the exit listener nulling the instance field could skip the SIGKILL escalation; also idempotent, and `start()` now resets the killed flag so a stopped manager can restart (required for the future singleton).

No public-surface changes: constructor stays backward compatible (`new MoveliteManager(binary)` unchanged; adapter is an optional third parameter), `NodeProvider` contract unchanged (now declared explicitly via `implements`).

New 26-test unit suite (`src/node/__tests__/MoveliteManager.test.ts`) mirroring the LocalNodeManager fake-adapter patterns, plus a per-file coverage threshold in vitest.config.ts.

## Tier 2 gates (MoveliteManager is on the backend-selection list, §6.2)

- `pnpm vitest run --coverage` — **pass** (617 tests, 54 files; new per-file threshold enforced by the run's exit code).
- `MOVEHAT_EXPECT_BACKEND=movelite pnpm assert-backend` — **pass** (backend=movelite, boot 5511ms <= 20000ms).
- `MOVEHAT_EXPECT_BACKEND=movement-node pnpm assert-backend` — **pass** (backend=movement-node, boot 25440ms <= 90000ms; scope guard falls through correctly).
- `pnpm test:example` — **pass** (9 passing, 12s).
- `pnpm test:smoke` — **pass**.

## Empirical lifecycle verification (issue #353 DoD)

- **No-stop exit**: a script that `start()`s movelite and returns without `stop()` exits in ~1s total (no event-loop hang) and leaves `pgrep -f movelite` empty with port 8090 refusing connections — unref + exit hook both verified.
- **SIGINT mid-suite**: interrupted `pnpm test:example` 12s in (tests running); `pgrep -f movelite` empty and port 8090 free afterwards.

## Tier 3

N/A for this sub-PR (no template or packaging changes); rides the batch e2e per §6.3.

Closes #353
Tracks #341

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,21 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Fixed
+
+- `MoveliteManager` process lifecycle is now safe for long-lived use.
+  Interrupting a test run with Ctrl+C no longer leaves an orphaned movelite
+  holding port 8090 (a sync kill hook now runs on process exit and on
+  SIGINT/SIGTERM), an unstopped node no longer keeps the mocha event loop
+  alive (the child and its stdio pipes are unref'd once ready), a chatty
+  node can no longer block on the undrained 64KB pipe buffer (output is
+  drained through the standard verbosity filter — critical signals always
+  surface, chatter only with `-v`), and a node that dies during boot now
+  fails immediately with its exit code and recent stderr instead of burning
+  the full 15s readiness timeout on a generic message. A stopped manager
+  can also be started again (the internal killed flag was never reset).
+  Closes #353.
+
 ### Internal
 
 - The pre-publish gate (`scripts/pre-publish.sh`) no longer aborts on the
--- a/packages/movehat/src/node/LocalNodeManager.ts
+++ b/packages/movehat/src/node/LocalNodeManager.ts
@@ -11,12 +11,12 @@ import { logger, isVerbose, colors, symbols } from "../ui/index.js";
 import { withTimedSpinner, withSpinner } from "../ui/spinner.js";
 
 /**
- * Substrings that always surface from the movement subprocess regardless
- * of verbosity. These are signals the user must see to debug a stuck
+ * Substrings that always surface from a node subprocess regardless of
+ * verbosity. These are signals the user must see to debug a stuck
  * startup (panic, fatal, address-in-use). Tested in
  * __tests__/LocalNodeManager.test.ts to guard against silent regressions.
  */
-const CRITICAL_NODE_OUTPUT = /panic|fatal|address already in use|EADDRINUSE/i;
+export const CRITICAL_NODE_OUTPUT = /panic|fatal|address already in use|EADDRINUSE/i;
 
 export interface LocalNodeOptions {
   testDir?: string;           // Directory for node data (default: .movehat/local-node)
--- a/packages/movehat/src/node/MoveliteManager.ts
+++ b/packages/movehat/src/node/MoveliteManager.ts
@@ -15,6 +15,7 @@ import { logger, isVerbose, colors, symbols } from "../ui/index.js";
 
 const READY_TIMEOUT_MS = 15_000;
 const READY_POLL_INTERVAL_MS = 200;
+const READY_PROBE_TIMEOUT_MS = 2_000;
 const SIGKILL_GRACE_MS = 5_000;
 const MAX_STDERR_TAIL_LINES = 20;
 const MAX_STDERR_TAIL_LINE_LENGTH = 500;
@@ -167,7 +168,12 @@ export class MoveliteManager implements NodeProvider {
       // not mask a child that already died.
       if (exitInfo) throw this.buildBootFailure(exitInfo);
       try {
-        const res = await fetch(url);
+        // Per-probe timeout: without it, a wedged listener that accepts
+        // TCP but never responds would stall the loop far past the 15s
+        // deadline, which is only checked between iterations.
+        const res = await fetch(url, {
+          signal: AbortSignal.timeout(READY_PROBE_TIMEOUT_MS),
+        });
         if (res.ok && !exitInfo) return;
       } catch {
         // not ready yet
--- a/packages/movehat/src/node/MoveliteManager.ts
+++ b/packages/movehat/src/node/MoveliteManager.ts
@@ -1,20 +1,41 @@
-import { execSync, spawn, type ChildProcess } from "child_process";
+import { execSync } from "child_process";
 import { existsSync } from "fs";
 import { join } from "path";
 import { createRequire } from "module";
 import type { Account } from "@aptos-labs/ts-sdk";
-import type { LocalNodeInfo } from "./LocalNodeManager.js";
-import { logger } from "../ui/index.js";
+import { CRITICAL_NODE_OUTPUT, type LocalNodeInfo } from "./LocalNodeManager.js";
+import type { NodeProvider } from "./NodeProvider.js";
+import {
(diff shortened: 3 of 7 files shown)


gilbertsahumada (commented):
## §8 Self-review

Re-read the full diff with specific attention to the failure modes this PR exists to close (event-loop retention, orphaning, pipe blocking, boot TOCTOU) and to whether the fix introduces new instances of the same classes.

### 🔴 Blockers

None.

None found. The candidates I chased down:

- **Does unref'ing change behavior for a consumer that relied on movelite keeping the process alive?** Audited all consumers (`setupLocalTesting`, template root hook, backend probe): every one runs under mocha or an awaited script whose own work holds the loop; none relied on the node's handles. The fork server (the one long-serving process) uses `ForkServer`, not `MoveliteManager`.
- **Hook lifetime vs the SIGTERM grace window**: `stop()` deregisters only in `finally` after `exited` resolves, so a parent death inside the 5s window still reaps the child. Verified by reading the ordering, and the escalation test pins the SIGTERM-then-SIGKILL sequence.
- **Stale `exitInfo` masking a genuinely ready node**: the liveness flag only ever transitions null -> exited, and a child that exited can never serve the ready URL itself — a ready response with `exitInfo` set is by definition a foreign node, so refusing it is correct.
- **`process.on("exit")` listener accumulation**: register-on-start / deregister-on-(stop or self-exit) keeps live listeners equal to live nodes, and the port logic caps those at 2 per host. The hook-count test pins the balance.

### 🟢 Nits / observations

1. `start()` has no re-entrance guard (calling it twice without `stop()` spawns twice). Pre-existing, no current consumer does it, and the singleton PR (sub-PR 3 of #341) is where that guard belongs — deferred deliberately per its design notes.
2. The §9 verbose-passthrough `console.log`/`console.error` at MoveliteManager.ts:80/101 are the sanctioned subprocess-passthrough pattern (verbosity-gated, mirroring LocalNodeManager); the §9 grep audit shows no ungated calls.
3. The stderr handler tests the critical regex against the whole chunk but tails per line — same asymmetry LocalNodeManager has; kept identical on purpose (shared filter, shared expectations).

### Verified

Tier 2 gates all pass (details in PR body); both empirical lifecycle checks (no-stop prompt exit, SIGINT mid-suite) left zero orphans and port 8090 free. Pre-push hook (build + unit tests + example typecheck) passed on push.

gilbertsahumada (commented):
## Multi-agent adversarial review (/review)

Two independent adversarial passes over the PR diff: one attacking the lifecycle correctness (every hook-registry interleaving, stop/exited races, signal paths, unref semantics, import cycles), one attacking test fidelity via **mutation testing** (7 mutants of MoveliteManager.ts run against the suite).

### Verdict on the production code: no confirmed bugs

The lifecycle reviewer could not construct a breaking interleaving. Specifically verified sound: idempotent identity-guarded hook deregistration across every path (stop, child self-exit, start failure, both orderings of stop vs the exited continuation); concurrent/double stop; stop during boot; signal path (the `setImmediate(process.exit)` in the shared handler is a ref'd handle, so an all-unref'd process cannot die before it fires); no import cycles; `tsc` clean; only in-repo `SpawnedProcess` implementers updated.

### Findings

🟡 **Fixed in 1c2b838 — unbounded in-flight readiness fetch defeated the 15s promise.** `waitForReady`'s deadline is only checked between iterations and the probe fetch had no timeout, so a wedged listener that accepts TCP but never responds stalled `start()` to undici's ~300s default while the eventual error claimed "within 15000ms". Each probe now carries a 2s `AbortSignal.timeout`.

🟡 **Fixed in 1c2b838 — three verified test-suite escapes (mutation testing).** Each of these mutants passed the original 26/26:
- Deleting the foreign-node guard (`res.ok && !exitInfo` → `res.ok`) — the headline TOCTOU claim rested half on an untested branch. New test: a ready response resolved after the child dies must reject.
- Removing the stderr `isVerbose()` gate (unconditional raw passthrough — a §9 violation) — the stderr quiet-suppression and verbose paths had no tests. Both added, mirroring the sibling suite.
- `stdio: "pipe"` → `"ignore"` and `.on("data")` → `.once("data")` — spawn stdio mode and drain wiring were unpinned. Now pinned via `calls[0].stdio`, `listenerCount("data")`, and two-chunk verbose assertions.
Also fixed: `vi.useRealTimers()` moved into `afterEach` (a failed assertion in a fake-timer test previously leaked fake timers and cascaded into misleading timeouts downstream), and the 500-char stderr-line truncation gained a test. Both headline mutants re-run post-fix and now fail the suite; full run 621/621 with coverage thresholds green.

🟢 **Noted, deferred with rationale:**
- No `start()` re-entrance guard (LocalNodeManager has one): a direct-API double-start leaks the first child until process exit and pollutes `stderrTail`. No in-repo consumer double-starts; the guard belongs to the singleton sub-PR (3 of 3 for the parent issue), where re-entrance becomes a real path.
- Microtask-width residual window in the foreign-node defense: `exitInfo` is set via `.then`, so a fetch resolving ok in the same event-delivery window as the child's death can still slip through. Requires the already-pathological foreign-node scenario with event-granularity timing; the guard now under test covers the practical window.
- Theoretical early process exit during `stop()` if movelite ignores SIGTERM and nothing else holds the loop; unreproducible with the real binary (it handles SIGTERM — the normal-stop path depends on it).
- Pre-existing, untouched: a failed both-ports-busy `start()` leaves `this.port` incremented across retries.

### Verified solid by the test reviewer

Fake-timer/Date mechanics (fail-safe: a regression hangs rather than false-passes), hook-count identity and symmetry, fail-fast timing (600ms vs 15s genuinely pinned), SIGTERM→SIGKILL escalation sequence, unref-once-after-ready ordering, restart-after-stop, tail boundedness.

PR #360, opened 2026-07-09
fix: close local-network deployment races (#341 sub-PR 2)

gilbertsahumada:
## Summary

Closes the three in-process deploy races that become reachable once fixtures share a node implicitly (#341 sub-PR 3):

1. **build/ stomp**: `movement move build` writes in-place to `<packageDir>/build/` with the deployer address baked into the bytecode, and the SDK publish path (movelite) reads those bytes back from disk non-atomically — two overlapping same-package deploys could publish bytecode compiled for the other deploy's address.
2. **Idempotency TOCTOU**: two concurrent same-module deploys both passed the already-deployed check and double-published.
3. **MH_CLI_REDEPLOY env mutation**: the autoDeploy loop's set/restore of the process-global env var could strand it as `true` when fixtures interleave.

Fix: new internal `withKeyedLock` utility (`src/utils/keyedMutex.ts`, FIFO promise-tail chain per key); both write paths (`Publisher.deploy`, `codeObject.executeMovementMoveObject` — shared lock map, so they serialize against each other) wrap idempotency check + build + publish + record under the lock, keyed by `resolve(packageDir)`. The autoDeploy loop passes an explicit `redeploy: true` option threaded through `deployContract` → `Publisher.deploy` / `deployCodeObject`; the env var stays honored as a fallback for the CLI's `--redeploy` flag.

Documented limits (in the utility docstring): in-process only (two mocha processes can still stomp); same moduleName from two different packageDirs stays last-writer-wins on the record (user error).

## Tests

- `keyedMutex.test.ts` (new, 5 tests): FIFO per key, rejection isolation, distinct keys run concurrently, entry cleanup, return value.
- `deployContract.test.ts` (+4): build-stomp regression (concurrent same-dir deploys each publish their own marker artifacts), idempotency TOCTOU (exactly one publish, loser throws `ModuleAlreadyDeployedError`), `redeploy: true` bypass without env, `redeploy: false` wins over env.
- **Mutation-verified**: with `withKeyedLock` replaced by a pass-through, both regression tests fail; restored, 630/630 green.

## Tier 1 / Tier 2 (§6.2)

- [x] Unit suite: 630/630 passing (was 621; +9 new)
- [x] Tier 1 `pnpm check:example`: exit 0
- [x] `pnpm test:example`: 9 passing (15s)
- [x] `pnpm test:smoke`: exit 0
- [x] `MOVEHAT_EXPECT_BACKEND=movelite pnpm assert-backend`: passed (boot 5602ms)
- [x] `MOVEHAT_EXPECT_BACKEND=movement-node pnpm assert-backend`: passed (boot 24535ms)
- `pnpm test:e2e:quick`: N/A — no changes under `templates/scripts/**`
- §9 grep: 15 matches before and after (all pre-existing; zero new)

Closes #359. Tracks #341.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -23,6 +23,24 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Fixed
 
+- Concurrent deploys of the same Move package no longer race. The build
+  step writes in-place to `<packageDir>/build/` with the deployer address
+  baked into the bytecode, and the SDK publish path (movelite) reads those
+  bytes back from disk — so two overlapping deploys could publish bytecode
+  compiled for the other deploy's address, and two same-module deploys
+  could both pass the already-deployed check and double-publish. Deploys
+  now serialize per package directory through an in-process keyed lock
+  covering the idempotency check, build, publish, and deployment record —
+  in both write paths (`deployContract` and `deployCodeObject`, which
+  share the lock). The autoDeploy loop also no longer mutates the
+  process-global `MH_CLI_REDEPLOY` environment variable (interleaved
+  set/restore across concurrent fixtures could strand it as `true`);
+  `deployContract` and `deployCodeObject` instead accept an explicit
+  `redeploy` option, with the env var still honored as a fallback for the
+  CLI's `--redeploy` flag. New public option, so the next release is a
+  minor bump. Serial mocha hides these races today; they become reachable
+  with the implicit shared node planned in #341. Closes #359.
+
 - `MoveliteManager` process lifecycle is now safe for long-lived use.
   Interrupting a test run with Ctrl+C no longer leaves an orphaned movelite
   holding port 8090 (a sync kill hook now runs on process exit and on
--- a/packages/movehat/src/__tests__/deployContract.test.ts
+++ b/packages/movehat/src/__tests__/deployContract.test.ts
@@ -754,3 +754,230 @@ describe("Publisher — SDK publish path (movelite backend)", () => {
     ).rejects.toThrow(/Expected exactly one compiled package/);
   });
 });
+
+/**
+ * The build step writes in-place to <packageDir>/build/ with the deployer
+ * address baked into the bytecode, and publishViaSdk reads those bytes
+ * back from disk. Without per-package-dir serialization, two overlapping
+ * deploys publish each other's artifacts. These tests pin the lock.
+ */
+describe("Publisher — per-package-dir deploy serialization", () => {
+  const PKG = "dummy";
+  let tmpCwd: string;
+  let tmpHome: string;
+  let origCwd: string;
+  let origHome: string | undefined;
+  let origRedeploy: string | undefined;
+
+  beforeEach(() => {
+    tmpHome = mkdtempSync(join(tmpdir(), "movehat-lock-home-"));
+    tmpCwd = mkdtempSync(join(tmpdir(), "movehat-lock-cwd-"));
+
+    writeFileSync(
+      join(tmpCwd, "movehat.config.js"),
+      `export default {
+  defaultNetwork: "testnet",
+  networks: { testnet: { url: "https://testnet.movementnetwork.xyz/v1", chainId: "testnet" } }
+};
+`
+    );
+
+    const moveDir = join(tmpCwd, "move");
+    mkdirSync(join(moveDir, "sources"), { recursive: true });
+    writeFileSync(
+      join(moveDir, "Move.toml"),
+      `[package]\nname = "${PKG}"\nversion = "0.0.1"\n\n[addresses]\n`
+    );
+    writeFileSync(join(moveDir, "sources", "dummy.move"), "// intentionally empty\n");
+
+    origHome = process.env.HOME;
+    process.env.HOME = tmpHome;
+    origCwd = process.cwd();
+    process.chdir(tmpCwd);
+    origRedeploy = process.env.MH_CLI_REDEPLOY;
+    delete process.env.MH_CLI_REDEPLOY;
+  });
+
+  afterEach(() => {
+    try {
+      process.chdir(origCwd);
+    } finally {
+      if (origHome === undefined) delete process.env.HOME;
+      else process.env.HOME = origHome;
+      if (origRedeploy === undefined) delete process.env.MH_CLI_REDEPLOY;
+      else process.env.MH_CLI_REDEPLOY = origRedeploy;
+      if (existsSync(tmpHome)) rmSync(tmpHome, { recursive: true, force: true });
+      if (existsSync(tmpCwd)) rmSync(tmpCwd, { recursive: true, force: true });
+    }
+  });
+
+  function makeMockAptos(hash: string) {
+    const publishPackageTransaction = vi.fn(async (args: unknown) => ({ __tx: true, args }));
+    const aptos = {
+      publishPackageTransaction,
+      transaction: { sign: vi.fn(() => ({})), submit: { simple: vi.fn(async () => ({ hash })) } },
+      waitForTransaction: vi.fn(async () => ({ success: true, hash })),
+    } as unknown as Aptos;
+    return { aptos, publishPackageTransaction };
+  }
+
+  /**
+   * Build adapter that actually writes marker artifacts into build/ (the
+   * way the real CLI bakes the deployer address into the bytecode), then
+   * yields long enough for a concurrent unserialized build to stomp them.
+   */
+  function makeMarkingAdapter(marker: number): ChildProcessAdapter {
+    return {
+      async run(input) {
+        if (input.args[1] !== "build") {
+          throw new Error(`SDK path must not invoke the CLI for: ${input.args.join(" ")}`);
+        }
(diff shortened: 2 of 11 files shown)


gilbertsahumada (commented):
## Independent adversarial review (not the author)

Scope: full PR diff plus surrounding code in the checkout (`Publisher.ts`, `codeObject.ts`, `Harness.ts`, `script.ts`, `commands/compile.ts`, `cli.ts`, `setupLocalTesting.ts`, type surfaces, CHANGELOG). Verdict up front: the lock design is correct, the regression tests are load-bearing (mutation-verified below), and I found no blockers. One 🟡 on the type surface, a handful of 🟢.

### Verified independently (ran/checked myself)

- `pnpm vitest run`: **630/630 passing** — matches the PR claim.
- **Mutation check re-run**: replaced the `withKeyedLock` body with `return fn();` (backup copy, not git) → both regression tests fail exactly as claimed (build-stomp publishes `2` markers instead of `1` distinct set; TOCTOU yields 2 fulfilled instead of 1+`ModuleAlreadyDeployedError`). Restored, suite green again, working tree clean.
- **keyedMutex edge probe** (standalone script): (a) a rejecting `fn` whose caller *never* attaches a handler produces **zero** `unhandledRejection` events — the tail's synchronously-attached handlers cover `result` before the caller can be late; (b) `void tail.finally(...)` can't itself go unhandled (`tail` is pre-swallowed, the callback is throw-free `Map` ops); (c) after a 500-call burst across 7 keys with mixed rejections, the map is **empty** (identity-checked cleanup works); (d) FIFO holds with a 3-deep queue.
- Tier 1 `pnpm check:example`: exit 0.
- §9 grep: 15 matches, all pre-existing — matches the claim. No new raw `console.*` in `src/`.
- Grep for `MH_CLI_REDEPLOY` across `src/`: only `cli.ts` (writer, `--redeploy` preAction hook intact) and the two `?? env` fallback readers. Nothing else observed the env var during autoDeploy, so removing the set/restore loses no behavior. The `??` (not `||`) correctly lets explicit `redeploy: false` beat the env var, and both directions are pinned by tests.
- **Lock coverage sweep** for other `<dir>/build/` writers: `commands/compile.ts` (`movehat compile`) also builds in place but is only reachable from the CLI action — single-flow, no in-process deploy concurrency, and cross-process is documented out of scope in the utility docstring. `harness/script.ts` (`run-script`) compiles the script inline and does not touch `<packageDir>/build/`. `Publisher.publishViaCli`'s internal rebuild happens inside the lock. No uncovered in-process write path found.
- **Test discrimination**: the build-stomp test uses the *same dir with different moduleNames* (`stomp_a`/`stomp_b`), so a lock wrongly keyed by moduleName would fail it; a lock that released before `saveDeployment` would let the TOCTOU test double-publish. Both failure modes are actually detected.
- `Harness.deployCodeObject` forwards `options` wholesale (`deployCodeObject(this.runtime, options)`), and `codeObject.deployCodeObject`'s field-by-field mapping into `ExecuteOptions` includes `redeploy: options.redeploy` — nothing dropped on the harness path.
- MAINTENANCE.md: "new features, new exports" → minor. The CHANGELOG's minor-bump note is consistent.
- Repo comment policy: no issue/PR/history refs in any new source comment (keyedMutex.ts, Publisher.ts, codeObject.ts, tests). Refs live only in CHANGELOG/PR body, which is the right place.
- Issue #359 exists, is open, and describes exactly these three races. CHANGELOG entry sits under `### Fixed` in `[Unreleased]` with `Closes #359` and 0.2.x-style prose density.

### Taken on faith (not re-run)

`pnpm test:example`, `pnpm test:smoke`, both `assert-backend` modes — reported in the PR body; not re-executed here.

### Findings

🟡 **`UpgradeCodeObjectOptions` now advertises a `redeploy` knob that is a silent no-op.** `UpgradeCodeObjectOptions extends DeployCodeObjectOptions` (`src/types/harness.ts:88`), so upgrade callers inherit `redeploy?: boolean` with a docstring promising "Skip the already-deployed check and overwrite the deployment record." But `upgradeCodeObject` doesn't forward it, and even if it did, `forceRedeploy` is only consulted under `checkIdempotency: true` — upgrade always overwrites. A user passing `redeploy: false` to `upgradeCodeObject` expecting a refusal gets an overwrite. Fix is ≤5 LoC (a one-line "ignored for upgrade — upgrade always overwrites" note on `UpgradeCodeObjectOptions`, or `Omit<DeployCodeObjectOptions, "redeploy">`), which per §8's defer threshold means it should land in this PR rather than a follow-up.

🟢 **Lock key uses `resolve()`, not `realpathSync()`.** Two symlink aliases of the same packageDir get different lock keys and won't serialize. Contrived in practice (packageDir comes from config or an explicit option, and fixtures share the literal path), and `realpath` adds an ENOENT failure mode — fine to leave, but worth a word in the docstring if sub-PR 3 ever derives packageDir from more than one source.

🟢 **Error-ordering behavior change**: `validatePathSafety` now runs before the already-deployed check (it must — the lock key needs `safeDir`). A caller with a bad `packageDir` *and* an existing record now gets a path-safety error where it previously got `ModuleAlreadyDeployedError`. No test or CLI flow depends on the old order as far as I can find; noting it so it's a decision, not an accident.

🟢 **Unused destructured `account` in `Publisher.deploy`** (`Publisher.ts:63`): after the split, `deploy()` only uses `moduleName` and `config`; `account` is consumed in `deployLocked`. Harmless (`noUnusedLocals` is off) but trims to `const { moduleName, config } = input;`.

🟢 **Redundant `dir` parameter threading**: `deployLocked(input, dir, safeDir)` / `executeMovementMoveObjectLocked(opts, dir, safeDir)` pass `dir` only for log strings; the locked functions could recompute `input.packageDir || config.moveDir` in one line. Marginal — keep if you prefer the explicitness.

### Adversarial angles checked and cleared

- FIFO fairness and no-await-between `tails.get`/`tails.set` (synchronous, so arrival order is chain order).
- Reentrancy deadlock is documented in the docstring.
- Documented limits (in-process only; same moduleName from two dirs stays last-writer-wins) match what the code actually does — the lock is keyed by dir, and the record key is (network, moduleName), so the cross-dir same-name race genuinely remains and is honestly labeled user error.
- `setupLocalTesting`'s per-module try/log/rethrow behavior is preserved verbatim; only the env choreography was removed.
- `templates/types/movehat.d.ts` gained `redeploy` on `deployContract` options, consistent with the runtime type (`sdkPublish`/`adapter` remain deliberately unexposed to user templates, pre-existing).

### Conclusion

No 🔴. The one 🟡 is a small type-surface honesty fix that per §8 belongs on this branch. Everything load-bearing in the PR body reproduced when I ran it myself.

gilbertsahumada:
## Resolution of review findings

- 🟡 `UpgradeCodeObjectOptions` inherited a silent no-op `redeploy` — **fixed** in 97a3fc9: `extends Omit<DeployCodeObjectOptions, "redeploy">`. Upgrades always overwrite, so the option is meaningless there; excluding it makes passing it a compile error instead of a silent no-op.
- 🟢 Unused destructured `account` in `Publisher.deploy` after the split — **fixed** in the same commit (dead code introduced by this PR, so cleaned per §3).
- 🟢 Lock key uses `resolve()` not `realpathSync()` — **deferred/kept**: symlink-aliased packageDirs pointing at the same package are contrived; `realpathSync` adds a syscall + throw-on-missing-path semantics for no realistic win. The docstring already frames the lock as best-effort in-process.
- 🟢 `validatePathSafety` now precedes the already-deployed check — **deferred/accepted**: error-ordering change verified by the reviewer to have no dependents; failing fast on a bad path before consulting records is the more sensible order.
- 🟢 `dir` parameter threaded alongside `safeDir` — **deferred/kept**: `dir` preserves the user-supplied (possibly relative) path for log messages, matching the pre-existing log output exactly.

Post-fix verification: tsc clean, 630/630 unit tests, Tier 1 `check:example` exit 0.

gilbertsahumada (commented):
## §8 Self-Review

**Re-read of the full diff after the independent review round.**

🔴 none.

🟡 none remaining — the one 🟡 (inherited no-op `redeploy` on upgrade options) is fixed in 97a3fc9.

🟢 observations from the re-read:
- The `deployLocked` / `executeMovementMoveObjectLocked` split keeps the diff reviewable (no 130-line re-indent) at the cost of two extra private function signatures; acceptable trade.
- The stomp regression test relies on `setTimeout(30)` inside the fake build to force overlap — it is deterministic in the failing direction (verified by mutation: lock removed → both concurrency tests fail 100%), and the lock direction has no timing dependence at all (strict serialization).
- `MH_CLI_REDEPLOY` handling now has three layers (option > env > default false); the precedence is pinned by the "redeploy: false wins over MH_CLI_REDEPLOY=true" test.

**Checklist:**
- [x] Independent agent review posted on the PR (0 🔴 / 1 🟡 / 4 🟢) + resolution comment
- [x] Mutation-verified regression tests (independently re-verified by the reviewer)
- [x] Tier 1 + all four Tier 2 gates green (results in PR body); e2e:quick N/A
- [x] CHANGELOG `[Unreleased] ### Fixed` entry, Closes #359, Tracks #341
- [x] §9 grep: zero new matches

Ready to merge pending checks.

